PR #5, opened 2022-02-01
Plugins initalizer

dvladir:
Move loading and registering plugins from bootstrap.js to APP_INITIALIZE phase


src/app/plugins-initializer.ts
```diff
@@ -0,0 +1,69 @@
...
+    document.querySelector('head')!.appendChild(script);
+  });
+
+const loadSinglePlugin = async (pluginDefinition: PluginDefinition): Promise<string[]> => {
+  try {
+    const scriptsLoad = pluginDefinition.scripts.map((script) => loadScript(script));
+    await Promise.all(scriptsLoad);
+  } catch (e) {
+    console.log(`Module(s) ${pluginDefinition.angularModules.join(', ')} load fail`);
+    console.log(e);
```
neogucky, line 31: I would suggest to use the second parameter of console.log when possible:

`console.log(`Module(s) ${pluginDefinition.angularModules.join(', ')} load fail`, e);`

neogucky (approved):
I left feedback to make console.logs cleaner but this is fine for now, great commit. 

As a sidenode, you included heavy changes to index.html mainly due to the formatting. It is important that we format our files nicely (and often the IDE does it automatically) but please try to commit formatting changes in seperate commits since it is dificult to see the important changes when the whole file is shown in highlight.

PR #6, opened 2022-02-07
Handle errors for plugins loading

dvladir:
1. I've add the folder src/plugins and put it to gitignore. It should affect the build, but it will help to test plugins locally. So I decided to include it in this commit also.
2. js/bootstrap.js was removed in previous PR, but I forgot to delete it's script inclusion. Now the inclusion has been deleted also.
3. onerror callback was added to plugin's script load to avoid the application start failure if one of the plugins is not available.

The plugins load became more stable. For example if `monitoringdashboard` wont be available, user won't see the tab monitoring, but the whole app will continue to work. But there are some cases, when error handling on scripts load won't help.
Some plugins are depend on each other. For example `monitoringdashboard` depends on `notifications`, and we can not analyze plugins dependencies during the load. It means that if `notifications` scripts won't be available, `monitoringdashboard` won't be able to start properly and it will brake the application.

### DIFF
--- a/src/app/plugins-initializer.ts
+++ b/src/app/plugins-initializer.ts
@@ -8,17 +8,19 @@ export interface PluginDefinition {
 }
 
 const loadScript = (fileName: string): Promise<unknown> =>
-  new Promise<unknown>((resolve) => {
-    const callback = () => {
+  new Promise<unknown>((resolve, reject) => {
+    const callbackSuccess = () => {
       console.log(`Loaded plugin script ${fileName}`);
       resolve(true);
     };
+    const callbackError = (err: any) => reject(err);
     const script = document.createElement('script');
     script.setAttribute('type', 'text/javascript');
     script.setAttribute('src', fileName);
     //@ts-ignore
-    script.onreadystatechange = callback;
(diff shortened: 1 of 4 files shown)


src/app/plugins-initializer.ts
```diff
@@ -27,8 +29,7 @@ const loadSinglePlugin = async (pluginDefinition: PluginDefinition): Promise<str
     const scriptsLoad = pluginDefinition.scripts.map((script) => loadScript(script));
     await Promise.all(scriptsLoad);
   } catch (e) {
-    console.log(`Module(s) ${pluginDefinition.angularModules.join(', ')} load fail`);
-    console.log(e);
+    console.log(`Module(s) ${pluginDefinition.angularModules.join(', ')} load fail`, e);
```
neogucky, line 32: We should use console.error here, because this is an error that should never occure for users and if it does, needs to be easy to detect.

neogucky (changes requested):
Makes sense, the small change should be included, then we can merge it into master

neogucky (approved):

PR #7, opened 2022-02-14
Ability to load hybrid plugins was added

dvladir:


src/app/plugins-initializer/plugins-initializer.ts
```diff
@@ -0,0 +1,64 @@
+import { APP_INITIALIZER, Compiler, FactoryProvider, Injector } from '@angular/core';
+import { A1PluginDefinition } from './shared/a1-plugin-definition';
```
neogucky, line 2: I missed that we comitted this to step-frontend, lets rename the plugin classes to something meaningful

src/app/plugins-initializer/plugins-initializer.ts
```diff
@@ -0,0 +1,64 @@
...
+import { A2PluginDefinition } from './shared/a2-plugin-definition';
+import { registerA1Plugins } from './register-a1-plugins';
+import { registerA2Plugins } from './register-a2-plugins';
+
+const OVERRIDE_PLUGINS = new Map<string, string>([['multitenancy', 'multitenancy/remoteEntry.js']]);
```
neogucky, line 7: Great, I think this is the right approach until we modified the BE! Maybe add a comment 
//TODO: modify in BE and remove override
just to make sure we don't forget

neogucky (changes requested):
We should rename the plugin classes to something other than PluginA1

PR #18, opened 2022-03-29
Deploy main app as npm library dependency

dvladir:


neogucky:
@dvladir After I added small changes to master (and then branched from master) there is a conflict, can you fix it?

PR #24, opened 2022-04-26
Vad/plans table

dvladir:


projects/step-core/src/lib/services/auth.service.ts
```diff
@@ -31,6 +31,10 @@ export class AuthService {
 
   private _serviceContext: { conf?: ConfigDto } = {};
 
+  private _context$ = new BehaviorSubject<AuthContext | undefined>(undefined);
+
+  readonly context$ = this._context$.asObservable();
```
neogucky, line 36: I generally like the idea to use $-suffix for observables. But:

1. isn't it incorrect for the BehaviorSubject?
2. we should discuss if we want to use it consistently everywhere

dvladir: @neogucky 

1) I think it is correct, because `BehaviourSubject` extends from `Subject`, and `Subject` extends from `Observable`.
So all of them are rxjs streams.
2) Personally I wish to use such notation for observable and subjects, because it will be easier to spot them in the code.

PR #32, opened 2022-06-07
SED-1087: Execution list.

dvladir:
- The core table was extended - now it allows to render custom search elements.
- For execution status and distribution bar - for now the same approach and the same colors were used as in original angularJS directives. In future these components should be redesigned to have a better look with material-design.
- entity-icon and report-node-icon directives were not migrated. We need to decide first what kind of material icon for each entity we should use.


projects/step-core/src/lib/modules/table/components/table/table.component.html
```diff
@@ -7,9 +7,14 @@
...
-      <mat-form-field *ngIf="searchCol.searchName">
-        <input matInput type="text" (change)="onSearch(searchCol.searchName, $event)" />
-      </mat-form-field>
+      <ng-template #defaultSearchRender>
+        <mat-form-field>
+          <input matInput type="text" (change)="onSearch(searchCol.searchName!, $event)" />
+        </mat-form-field>
+      </ng-template>
+      <ng-container *ngIf="searchCol.searchName">
+        <ng-container *ngIf="!!searchCol.template; then searchCol.template!; else defaultSearchRender"> </ng-container>
```
neogucky, line 16: I don't really get how the `then searchCol.template!` works here, can you explain it to me?

neogucky: In my understanding the then / else directives within ngIf reference layouts declared with #layoutName. Like you do with defaultSearchRender in the same place. Where is `searchCol.template` specified in the view or is it doing something else in this case?

dvladir: It is taken from `SearchColDirective` in case if it has a `SearchCellDefDirective` defined as it's content.
searchCol.template is filled here:
https://github.com/exense/step-frontend/pull/32/files#diff-32279eb8c316e816b9a55d3142427d11701e963a1ab2f02a3055328b08e4b094R110

This template appears when the column is defined with following syntax:
```
  <ng-container matColumnDef="myColumn" stepSearchCol>
    <div *stepSearchCellDef>SEARCH CELL CONTENT</div>
    <th mat-header-cell *matHeaderCellDef>My column header</th>
    <td mat-cell *matCellDef="let element">{{ element?.someField }}</td>
  </ng-container>

```
The `*stepSearchCellDef` makes the trick. `stepSearchCol` see that it have a stepSearchCellDef as child and take it's template.
`*` notation is as shortcut for:
```
<ng-template [stepSearchCellDef]>
  <div>SEARCH CELL CONTENT</div>
</ng-template>
```

neogucky (approved):

chillbirdo (approved):

PR #33, opened 2022-06-08
Feature generate openapi client

chillbirdo:
- minor issue with Api Specs: https://trello.com/c/O5VoMq3h/75-commuinicate-openapi-spec-issue
- please create follow-up tickets: in the new part of the application, all manually created Api-Calls as well as all DTOs (in `projects/step-core/src/lib/domain`) should be replaced by the generated Services and DTOs
- confluence documentation will follow


chillbirdo:
- please note that generated code should never go through prettify, since that would make it impossible to view what fields have changed after regenerating the code. It would be useful to exclude the `generated` directory from prettify

dvladir:
Is it possible to configure the autogeneration to use the `kebab-case` for file names?

chillbirdo:
I compared 3 generators:
1) https://github.com/drwpow/openapi-typescript : does only create one ts file which is ugly and makes it difficult to understand what has really changes
2) https://github.com/OpenAPITools/openapi-generator : most mature, possibly supports distinct case types, but written in Java, so JRE required
3) https://github.com/ferdikoomen/openapi-typescript-codegen : my choice, TS and clean code, even that there are not many configuration options

neogucky (dismissed):

dvladir (approved):

neogucky:
Sounds good to me! I guess it should work on windows, if not when can look into that if we encounter the problem, there will always be a Linux machine around. 

Is it from your side ready to be merged?

neogucky (dismissed):

neogucky (approved):

PR #38, opened 2022-06-16
Sed 1102 adding plan list leftovers

neogucky:


neogucky:
@dvladir I have some questions on how to downgrade the entity-module so we can use it with A1. Namely I would want to use it in the multitenancy plugin. If it's easy just drop a message else lets have a call.

dvladir (changes requested):

dvladir:
> @dvladir I have some questions on how to downgrade the entity-module so we can use it with A1. Namely I would want to use it in the multitenancy plugin. If it's easy just drop a message else lets have a call.

As example for service downgrade, you can look at `auth.service` in step-core
As example for component downgrade 'tenant-selection.component' in the multitnenancy

Plus in your case, I think all the things that you want to downgrade, should be registered in the angularjs mode.
I think that `stepCoreJs`, that is defined in step-core/lib/angularjs/module.ts will be suitable for this.
Then you will also have to add this module to multitenancy dependencies.

projects/step-core/src/lib/modules/entity/components/entity-icon/entity-icon.component.ts
```diff
@@ -0,0 +1,37 @@
...
+export class EntityIconComponent {
+  @Input() entityName: string = '';
+  @Input() entity!: Entity;
+
+  icon?: string;
+  tooltip?: string;
+
+  constructor(private entityScopeResolver: EntityScopeResolver, private entityRegistry: EntityRegistry) {}
+
+  ngOnInit(): void {
```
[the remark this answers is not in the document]
neogucky: You are right, this will become relevant if we use pagination that might reuse container or similar

chillbirdo:
I noticed that there is an old directive with the same selector defined here:
https://github.com/exense/step-frontend/blob/59e321688f0231909b5625e0ca9f41311c904418/projects/step-frontend/src/lib/angularjs/js/controllers/entities.js#L85
used here:
https://github.com/exense/step-frontend/blob/59e321688f0231909b5625e0ca9f41311c904418/projects/step-frontend/src/lib/angularjs/partials/screenconfiguration/screenInputSelectionTable.html#L12

neogucky:
> I noticed that there is an old directive with the same selector defined here:
> 
> https://github.com/exense/step-frontend/blob/59e321688f0231909b5625e0ca9f41311c904418/projects/step-frontend/src/lib/angularjs/js/controllers/entities.js#L85
> 
> 
> used here:
> https://github.com/exense/step-frontend/blob/59e321688f0231909b5625e0ca9f41311c904418/projects/step-frontend/src/lib/angularjs/partials/screenconfiguration/screenInputSelectionTable.html#L12

That is correct, my component will replace that one either as part of this change or later on

projects/step-core/src/lib/modules/entity/components/entity-icon/entity-icon.component.ts
```diff
@@ -0,0 +1,47 @@
...
+  @Input() entityName?: string;
+  @Input() entity!: Entity;
+
+  icon?: string;
+  tooltip?: string;
+
+  constructor(private entityScopeResolver: EntityScopeResolver, private entityRegistry: EntityRegistry) {}
+
+  ngOnChanges(): void {
+    this.icon = 'adjust';
```
dvladir, line 24: Still required some fixes. `ngOnChanges` is invoked on every change detection cycle. To prevent redundant code executions, we need to check the required input parameters were really changed. Usually I use the following approach:
```
// Define additional method with the main logic of change's reaction. 
// As parameters it should accepts values that reaction is depends on
// In current case there are entity and entityName
// Both parameter are optional
private updateIconAndTooltip(entity?: Entity, entityName?: string): void {

   // If paramter not set, take value from the field
  entity = entity || this.entity;
  entityName = entityName || this.eneityName;

  // Do other stuff, which is based on parameter'values
  ....
}

// In ngOnChanges analyze which parameters were changed
// and invoke the reaction method
ngOnChanges(changes: SimpleChanges): void {

  let entity: Entity | undefined = undefined;
  let entityName: string | undefined = unedfined;

  const cEntity = changes.entity;
  const cEntityName = changes.entityName;

  if (cEntity?.previousValue !== cEntity?.currentValue) {
     entity = cEntity?.currentValue;
  }  

  if (cEntityName?.previousValue !== cEnityName?.currentValue) {
     entityName = cEntityName?.currentValue;
  }

  // entity and entityName variables will be undefined, in case if there is no changes for them 
  // In that case no reaction will be invoked.

  // Otherwise it will be invoked even one of them will be changed.
  if (entity || entityName) {

    // In case if entity or entityName will be empty
    // updateIconAndTooltip will take acutal value from component's field.
    this.updateIconAndTooltip(entity, entityName);  
  }

}
```
I hope the idea is clear :-) Feel free to modify it, if you see how it can be written in more optimal and clear way.

neogucky: I ended up doing something similar. I noticed that for some angularjs components implementing the new ng2+ component for the initial value previousValue and currentValue where the same thus skipping the update function, so I also looked for changes.xyz.firstChange.

neogucky:
@neogucky Reminder: when merging also merge the other branches

chillbirdo:
Plan List still does not reload after importing a plan

PR #39, opened 2022-06-20
Vad/sed 1097

dvladir:
- Typescript configuration to solve the paths properly in the IDE
- Shared mapping improvements


neogucky (approved):

chillbirdo:
looks good to me and more importantly: i tried it an it works like a charm!!

PR #40, opened 2022-06-21
 SED-1047 refactored resourcelist

chillbirdo:


chillbirdo:
sry forgot to use `SED-1047` in first commit

projects/step-frontend/src/lib/modules/context-menu/resources-list/resources-list.component.html
```diff
@@ -0,0 +1,59 @@
...
+    >
+      New Resource
+    </button>
+  </step-btn-group>
+</section>
+<step-table [dataSource]="dataSource" matSort [matSortDisableClear]="true">
+  <ng-container matColumnDef="name" stepSearchCol>
+    <th mat-header-cell *matHeaderCellDef mat-sort-header>Name</th>
+    <td mat-cell *matCellDef="let element">
+      <a href="#" (click)="$event.preventDefault(); editResource(element)">{{ element.attributes.name }}</a>
```
dvladir, line 18: As alternative, we can prevent link navigation if the event handler will return `false`.
So the `editResouce` itself can return false, or otherwise if we don't want to change method's return type, we can write it as
```
<a href="#" (click)="editResource(element);false">
{{eleement.attributes.name}}
</a>
```

chillbirdo: Thank Vlad for the hint! Although both of the suggested alternatives have less code, I find the explicit version with `preventDefault()` still easiest to understand. E.g. a new coder in the project works on this code and wonders why this method or expression returns `false`, he wants to clean up and refactors it. In contrast, `preventDefault()` gives him a clue why this piece of code is here.

projects/step-core/src/lib/modules/table/shared/table-remote-data-source.ts
```diff
@@ -107,7 +110,8 @@ export class TableRemoteDataSource<T> implements TableDataSource<T> {
   constructor(
     private _tableId: string,
     private _rest: TableRestService,
-    private _requestColumnsMap: { [key: string]: string }
+    private _requestColumnsMap: { [key: string]: string },
+    private _filter?: string
```
dvladir, line 114: I've just checked and noticed that some tables have this filter as dynamic value.
For example table in the `searchResourceDialog.html`

I would suggest to add filter a input parameter to `table.component`. And link this parameter with the subject
```
private _filter$ = new BehaviorSubject<string>(undefined);

get filter(): string {
  return this._filter$.value;
}

@Input() set filter(value: string) {
   if (value === this.filter) {
     return;
   }
  this._filter$.next(value);
}

```

Then add this subject to the combineLatest with page$,  sort$ and search$
```
combineLatest([page$, sort$, this._search$, this._filter$])
      .pipe(takeUntil(this._terminator$))
      .subscribe(([page, sort, search, filter]) => tableDataSource.getTableData(page, sort, search, filter));
```

And as final step extend `getTableData` method signature with new parameter.
For TableLocalDataSource the filter parameter will be ignored

dvladir (changes requested):

projects/step-core/src/lib/modules/table/shared/table-remote-data-source.ts
chillbirdo: Done! Yes indeed much more convenient and dynamic and the old `st-table` directive had the same feature as well

dvladir: @chillbirdo Thank you. Looks nice!
Just a few tiny improvements:

- I guess that _filter argument can be removed from TableRemoteDataSource constructor. I don't see that it's used anymore
- Also in `table.component.ts` can you complete the `_filter$` subject in the `ngOnDestroy` method as it is done for `_search$`

chillbirdo: Yes, subjects need to be completed, I overlooked that, thanks! I'll check the `_filter` argument too

neogucky:
> sry forgot to mention `SED-1047` in first commit

This is actually not a problem anymore since we will anyway squash ist

projects/step-frontend/src/lib/modules/context-menu/resources-list/resources-list.component.html
neogucky: I would prefere  `[routerLink]=""` instead of `href="#"`

projects/step-core/src/lib/modules/table/shared/table-remote-data-source.ts
chillbirdo: done!

dvladir (approved):

projects/step-frontend/src/lib/modules/context-menu/resources-list/resources-list.component.html
neogucky: That is weird, routerLink should work for Angular 2+, in AngularJS you had to do  `href="#" ng-click="$event.preventDefault();`

dvladir: @neogucky As i remember, we decided not to migrate the routing for now.
[routerLink] make sense only in that case, if you want to make a navigation through the Angualr 2+ route
I don't see anything critical in `<a href="#">` in Angular 2+, when it plays the role of button that does some action. 
The only thing the we need to do in such cases, to prevent the original even behavior

neogucky (approved):

PR #41, opened 2022-06-22
Call API from single module

chillbirdo:
Dear fellows,

on my way to solve https://exense.atlassian.net/browse/SED-1105 , I got tempted to refactor some more of the structure of the project, but wasn't sure whether you would agree with my ideas. So I created a demo-branch where I changed some parts of the structure and applied it to `plan-list.component.ts`. I you are fine with that I would apply the changes to SED-1105

The problem I see is the following:

## Distinct ways and places to call the API
At the moment, from our new components, the API is called in 4 distrinct ways: using `httpClient`, by creating a `TableRemoteDataSource`, using our generated API-services, using legacy code from the AngularJS Modal Controllers.

In my opinion it would be much cleaner to deal with the API from within a single module (the `StepClientModule`) and use as much of the generated code as possible, i.e. never address an endpoint in the code, use generated dtos when possible. **I moved the code that invokes the Table endpoints of the API to one unified client module**. I suggest to stick with the following guideline when writing our new components:

- only import one service that deals with the backend. (no `httpClient`, no `TableRestService`, neither in the component, nor in the `-dialogs.service.ts`) In my example it is `AugmentedPlansService`, which augments the generated `PlansService` to fetch a `Plan` dto from the table interface.

What do you think about it? Since I moved some files around it might be useful to checkout the branch in order to see my suggestions.

Thanks and best regards,
Gilbert

### DIFF
--- a/projects/step-frontend/src/lib/modules/plan/components/plan-list/plan-list.component.ts
+++ b/projects/step-frontend/src/lib/modules/plan/components/plan-list/plan-list.component.ts
@@ -7,7 +7,6 @@ import {
   DialogsService,
   AugmentedPlansService,
   Plan,
-  TableRemoteDataSource,
 } from '@exense/step-core';
 import { PlanDialogsService } from '../../servies/plan-dialogs.service';
 import { ExportDialogsService } from '../../../_common/services/export-dialogs.service';
@@ -23,22 +22,18 @@ import { ILocationService } from 'angular';
   styleUrls: ['./plan-list.component.scss'],
 })
 export class PlanListComponent {
-  readonly dataSource: TableRemoteDataSource<Plan>;
-
   constructor(
-    private _augmentedPlansService: AugmentedPlansService,
+    readonly _plansApiService: AugmentedPlansService,
     private _dialogs: DialogsService,
     private _planDialogs: PlanDialogsService,
     private _exportDialogs: ExportDialogsService,
     private _importDialogs: ImportDialogsService,
     private _isUsedByDialogs: IsUsedByDialogsService,
     @Inject(AJS_LOCATION) private _location: ILocationService
-  ) {
-    this.dataSource = this._augmentedPlansService.getPlansTableDataSource();
-  }
+  ) {}
 
   addPlan(): void {
-    this._planDialogs.createPlan().subscribe((_) => this.dataSource.reload());
+    this._planDialogs.createPlan().subscribe((_) => this._plansApiService.reloadPlansTableDataSource());
   }
 
   editPlan(id: string): void {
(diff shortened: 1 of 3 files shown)


projects/step-frontend/src/lib/modules/plan/components/plan-list/plan-list.component.ts
```diff
@@ -53,16 +45,16 @@ export class PlanListComponent {
...
   duplicatePlan(id: string): void {
-    this._httpClient
-      .get<any>(`rest/plans/${id}/clone`)
+    this._plansApiService
+      .clonePlan(id)
       .pipe(
-        map((clone) => {
-          clone.attributes.name += '_Copy';
+        map((clone: Plan) => {
+          clone['attributes']!['name']! += '_Copy';
```
chillbirdo, line 52: if we would disable `"noPropertyAccessFromIndexSignature": true` from `tsconfig.workspace.json` we could use the `Plan` type here (even without the explicit `clone: Plan` typing, since `clonePlan()` returns a `Plan` anyways). Just asking: do we use this compilerOption intentionally?

neogucky: @chillbirdo I guess we have it on to be more strict, but for me it would be fine to set it to false. In the very least I would write known property still with the dot syntax ( clone.attributes!['name']! += '_Copy';) If this is the only possible occurence we could also define a type for attributes.

projects/step-frontend/src/lib/modules/plan/components/plan-list/plan-list.component.ts
```diff
@@ -53,16 +45,16 @@ export class PlanListComponent {
...
           return clone;
         }),
-        switchMap((clone) => this._httpClient.post('rest/plans', clone))
+        switchMap((clone) => this._plansApiService.save4(clone))
       )
-      .subscribe((_) => this.dataSource.reload());
+      .subscribe((_) => this._plansApiService.reloadPlansTableDataSource());
```
neogucky, line 57: This did not work for me when I tested locally (creating new plan did not show new plan)

neogucky (commented):
Let's have a call tomorrow. I like having everything in one place. We should also think about what APIs could be merged and have a discussion with the BE.

projects/step-frontend/src/lib/modules/plan/components/plan-list/plan-list.component.ts
chillbirdo: thanks.
additionally, when importing a plan it does not update as well, but on master!

neogucky: Interesting, I never tried importing. Adding and duplicating updates on the master for me though.

neogucky (commented):
looks good

PR #42, opened 2022-06-27
All Timeseries Code (straight merge)

lucianbaciu:


projects/step-frontend/src/lib/modules/timeseries/ranger/ts-ranger.component.ts
```diff
@@ -0,0 +1,284 @@
...
+    }, 50);
+  }
+
+  createRanger() {
+    // @ts-ignore
+    let x0;
+    let lft0: number;
+    let wid0: number;
+    // @ts-ignore
+    const lftWid = { left: null, width: null };
```
neogucky, line 73: I picked this one as an example, I dislike abreveated varibles like "lft" or "el" as long as the variables do not become super long oterwise. The "lftWid" variable could be split into "width" and "left" and the nassembled when used like:

```
let select = (newLeft, newWidth) => {
      this.uplot.setSelect({left: newLeft, width: newWidth}, false);
    };
```

chillbirdo:
image

To reproduce: Tried it against `stepos-nhy.exense.ch` backend. I imported one plan from the sample project, executed it. So currenty we have one Execution in the nhy OS database. From the `Execution List` view I clicked on the link next `(?)`

projects/step-frontend/src/lib/modules/timeseries/ranger/ts-ranger.component.ts
lucianbaciu: These functions are not implemented by me, but "part" of the library. My scope was to have it up and running asap, that's why I left it the way it was. Same for ts-ignore. There are @ts-ignore just on the functions written by the library contributors. Of course we can clean all them up.

projects/step-frontend/src/lib/modules/timeseries/ranger/ts-ranger.component.ts
```diff
@@ -0,0 +1,284 @@
...
+      this.uplot.setSelect({ left, width, height }, false);
+      let xData = this.uplot.data[0];
+      let start = xData[0];
+      let end = xData[xData.length - 1];
+      this.onRangeChange.emit({ start, end });
+    }, 50);
+  }
+
+  createRanger() {
+    // @ts-ignore
```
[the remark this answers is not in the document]
lucianbaciu: No. it would be ideal to change them to ts.

lucianbaciu:
@chillbirdo - you need the backend configured to properly start to ingest data needed for this feature. Corner cases like "no data" are not handled yet, but will be soon.

chillbirdo:
@lucianbaciu good work! Charts look really good when there is data.
Indeed, "no data" is a corner case. Anyways, it throws an error and I guess it should be fixed before going into the master. If it is an urgent feature and the "no-data" fix takes too long, we would need to open a bug-issue. The decision remains to @neogucky

neogucky (dismissed):

neogucky (approved):

PR #43, opened 2022-06-27
Refactors Parameters table

chillbirdo:
Slightly changes orderBy behaviour of table component


projects/step-core/src/lib/modules/table/shared/table-remote-data-source.ts
```diff
@@ -67,16 +67,17 @@ const convertTableRequest = (req: TableRequest): TableRequestData => {
...
+  if (!!req.orderBy && req.columns.indexOf(req.orderBy.column) >= 0) {
+    orderColumnIndex = req.columns.indexOf(req.orderBy.column);
   }
+  result.columns[orderColumnIndex].orderable = true;
+  result.order = [
+    {
+      column: orderColumnIndex,
+      dir: !!req.orderBy?.order ? req.orderBy.order : 'asc',
+    },
+  ];
```
neogucky, line 80: This change looks like we could also replace  `: -1;`  with `: 0;`. I added a discussion point to Trello.

neogucky (changes requested):
You can check this branch here: https://stepos-sed-1040-refactor-parameters-table.exense.ch/#/root/parameters

The "new parameter" throws an exception

chillbirdo:
damn it worked until I shut down and restarted from workspace. next time I'm not going to push without a check!

chillbirdo:
works now!

PR #44, opened 2022-06-28
SED-1112 minor design rework

neogucky:
Changed design for tables, some actions required when migrating new tables (see PR)


neogucky:
https://stepos-sed-1112-minor-design-rework.exense.ch/

chillbirdo (approved):

PR #45, opened 2022-06-30
Sed 1063 scheduler table

chillbirdo:


projects/step-frontend/src/lib/modules/scheduler/execution-task-parameters-list/execution-task-parameters-list.component.ts
```diff
@@ -0,0 +1,141 @@
...
+  ) {
+    this.currentUserName = context.userName;
+  }
+
+  private loadTable(): void {
+    this.executionTaskParametersRequest$.next({});
+  }
+
+  executeParameter(executionTaskParameters: ExecutiontTaskParameters) {
+    //@ts-ignore
```
chillbirdo, line 85: The method of the generated `_schedulerService` could not be used in this case, becasue the API is returning a plain string instead of a JSON. That should be changed on API side.
Anyways, the way I worked around that is not ideal. In my opinion, the ideal solution would be to create an `AugmentedSchedulerService` like I've been showing in #41  , since this is code that belongs to the API-client. If eventually the API would return a JSON, we would just need to remove the augmentation.

projects/step-frontend/src/lib/modules/scheduler/execution-task-parameters-list/execution-task-parameters-list.component.html
```diff
@@ -0,0 +1,93 @@
...
+  <step-btn-group>
+    <button
+      type="button"
+      mat-raised-button
+      color="primary"
+      [disabled]="!('user-write' | hasRight | async)"
+      (click)="createParameter()"
+    >
+      New parameter
+    </button>
```
neogucky, line 11: We have a new way to do the primary add button as introduced in my table redesign PR.

`<button
      type="button"
      **mat-icon-button**
      **class="primary"**
      **matTooltip="new task"**
      **matTooltipPosition="below"**
      [disabled]="!('user-write' | hasRight | async)"
      (click)="createParameter()"
    >
      **<mat-icon>add</mat-icon>**
    </button>`

Take note that many properties changed

projects/step-frontend/src/lib/modules/scheduler/execution-task-parameters-list/execution-task-parameters-list.component.html
```diff
@@ -0,0 +1,93 @@
...
+  </step-btn-group>
+
+  <button
+    type="button"
+    mat-raised-button
```
neogucky, line 16: `    mat-raised-button`

projects/step-frontend/src/lib/modules/scheduler/execution-task-parameters-list/execution-task-parameters-list.component.scss
```diff
@@ -0,0 +1,47 @@
...
+.users-header {
+  margin-top: 1rem;
+  margin-bottom: 1rem;
+  display: flex;
+  justify-content: flex-end;
+
+  step-btn-group {
+    margin-left: 1rem;
+  }
+}
```
neogucky, line 47: This is not the users component, but because I saw this mistake multiple times I renamed the class from "COMPONENT-header" to just "header", we should do the same here. (Probably we can put much code somewhere global later on)

neogucky:
I saw that you wrote "parameter" a lot both in the code and in the labels. I don't really get where you have this from, since this table is a scheduler that contains scheduled tasks. Also try to re-use old labels, if one is badly named transparently add it as a comment to the PR. For example, I prefer: "Analyze performance in interactive session" instead of "See Statistics for parameter".

projects/step-frontend/src/lib/modules/scheduler/execution-task-parameters-list/execution-task-parameters-list.component.ts
```diff
@@ -0,0 +1,141 @@
...
+  readonly currentUserName: string;
+  readonly inProgress: boolean = false;
+
+  private executionTaskParametersRequest$ = new BehaviorSubject<any>({});
+  private executionTaskParameters$ = this.executionTaskParametersRequest$.pipe(
+    tap((_) => ((this as InProgress).inProgress = true)),
+    switchMap((_) => this._schedulerService.getScheduledExecutions()),
+    tap((_) => ((this as InProgress).inProgress = false)),
+    shareReplay(1)
+  );
```
neogucky, line 41: @chillbirdo 
I guess we use this to solve the reload issue, right?

I just discovered we have that one in all migrated tables, I don't like this code though. Can you check if there is a solution to nicely capsulate this in a service / inheritedServiceClass so we only call "service.reloadData()" or something in that direction?

chillbirdo: Absolutely, I was thinking the same, I actually did a note to myself here: https://github.com/exense/step-frontend/pull/41#issuecomment-1168715952

chillbirdo:
Thanks for the feedback, it is implemented now. I will take care about using the same wording as in the old version

neogucky (changes requested):
I noticed the following when reviewing:

- entity-icon missing (probably because the branch it depends on wasn't merged yet)
- Delete dialog still says "parameter"

Environment: https://stepos-sed-1063-scheduler-table.exense.ch/#/root/scheduler

When these are fixed it's great!

chillbirdo:
> I noticed the following when reviewing:
> 
> * entity-icon missing (probably because the branch it depends on wasn't merged yet)
> * plan link missing
> * Delete dialog still says "parameter"
> 
> Environment: https://stepos-sed-1063-scheduler-table.exense.ch/#/root/scheduler
> 
> When these are fixed it's great!

- entity-icon added
- plan link has already been there for tasks linked to a plan (as in the old version)
- delete dialog now says "task"

neogucky (approved):

PR #46, opened 2022-07-02
SED-1060 refactor table functions list

neogucky:


chillbirdo:
image

Recreate: try to duplicate a newly created keyword (like the ones I have just created). This error pops up, no duplication is shown to the user (although one is happening, it can be seen by pressing F5)

chillbirdo:
This legacy controller could be removed: https://github.com/exense/step-frontend/blob/58fe8083cd0402929260e202783e1c3b0776099e/projects/step-frontend/src/lib/angularjs/js/controllers/functions.js#L160

chillbirdo:
Just to check: The scope and description info (on mouseover the icons) is omitted intentionally?
image

neogucky:
> This legacy controller could be removed:
> 
> https://github.com/exense/step-frontend/blob/58fe8083cd0402929260e202783e1c3b0776099e/projects/step-frontend/src/lib/angularjs/js/controllers/functions.js#L160

No, other components still use this. After they are merged we can remove it safely.

neogucky:
> Just to check: The scope and description info (on mouseover the icons) is omitted intentionally? image image I guess you want to merge #38 first right?

Yes, we have lots of crossover dependencies of our current tickets and I still don't have a nice idea how to solve this. Probably we should do something iterative and merge as fast as possible and create some cleanup tickets. But you are right, we should create tickets for this and reference them in here so we don't forget them and it is easier to review.

neogucky:
Leftovers blocked by other tickets: https://exense.atlassian.net/browse/SED-1122

chillbirdo:
> > This legacy controller could be removed:
> > https://github.com/exense/step-frontend/blob/58fe8083cd0402929260e202783e1c3b0776099e/projects/step-frontend/src/lib/angularjs/js/controllers/functions.js#L160
> 
> No, other components still use this. After they are merged we can remove it safely.

For me to learn how the legacy code works: which components use it, I can not find a reference to it and everything I tested works fine without it.

neogucky:
> FunctionListCtrl

The mask list in step-enterprise-frontend (plugins\step-functions-plugins-pdftest\partials\masksList.html) uses it. In my IDE I usually search with ctrl + shift + f in both repositories for the controller name.

chillbirdo:
This is very helpful. I also search with `ctrl+shift+f` but I open step-frontend separately in my IDE so I do not overlook both projects! I hope that did not lead to errors already, I'll take care from now on

chillbirdo (dismissed):

PR #49, opened 2022-07-04
Sed 1102 adding plan list leftovers review1

chillbirdo:


neogucky (approved):
We really need to think about the  EntityRegistry.registerEntity. I'm also not 100% sure that removing this doesn't have any effects, so we should have an eye on it.

PR #50, opened 2022-07-04
Sed 1105 refactor refactored tables

chillbirdo:


neogucky (commented):
Looks good to me! I think we can take it out of draft.

neogucky (dismissed):

chillbirdo:
> Looks good to me! I think we can take it out of draft.

can me merged imo

chillbirdo:
I applied the following rules:
- all the code that communicates directly with the backend goes to the client module. the table component should only import a single service that deals with all the API communication
- all the code that deals with opening dialogs goes to the `[tableName]-dialogs.service`

projects/step-core/src/lib/client/augmented/services/augmented-executions-service.ts
```diff
@@ -0,0 +1,41 @@
...
+import { TableRestService } from '../../table/services/table-rest.service';
+import { BaseHttpRequest } from '../../generated/core/BaseHttpRequest';
+import { TableRemoteDataSource } from '../../../modules/table/shared/table-remote-data-source';
+import { Observable } from 'rxjs';
+
+@Injectable({ providedIn: 'root' })
+export class AugmentedExecutionsService extends ExecutionsService {
+  readonly EXECUTIONS_TABLE_ID = 'executions';
+
+  readonly dataSource = new TableRemoteDataSource(this.EXECUTIONS_TABLE_ID, this._tableRest, {
```
dvladir, line 12: Yes, it's nice idea to place remote datasources outside of the components. It will allow to reuse them.
But what it the benefit of mixing it with entity's api service?

If we are looking at it, like to an additional abstraction layer, then all entities should receive augmented version of api's service. All other parts of the application shouldn't use generated api services at all and do all interactions through the augmented variants.

Otherwise it brings duality, because if any other developer will decide to use the api, he can do it by original service injection or by the augmented service injection.

projects/step-core/src/lib/client/augmented/services/augmented-executions-service.ts
```diff
@@ -0,0 +1,41 @@
...
+    environment: 'executionParameters.customParameters.env',
+    status: 'status',
+    result: 'result',
+  });
+
+  constructor(override httpRequest: BaseHttpRequest, private _tableRest: TableRestService) {
+    super(httpRequest);
+  }
+
+  public getExecutionsTableDataSource(): TableRemoteDataSource<any> {
```
dvladir, line 26: This method is redundant, because dataSource is already public and readonly

projects/step-core/src/lib/client/augmented/services/augmented-executions-service.ts
```diff
@@ -0,0 +1,41 @@
...
+    return this.dataSource;
+  }
+
+  public reloadExecutionsTableDataSource(): void {
```
dvladir, line 30: This method is redundant, because reload can be accessed from the dataSource

projects/step-core/src/lib/client/augmented/services/augmented-executions-service.ts
```diff
@@ -0,0 +1,41 @@
...
+    this.dataSource.reload();
+  }
+
+  public getResultColumnItems(): Observable<string[]> {
```
dvladir, line 34: Should we specify public modifier?
By default if there is no private or protected modifier, typescript interprets it like public

projects/step-core/src/lib/client/augmented/services/augmented-executions-service.ts
chillbirdo: The idea is that the generated services can be extended if needed, e.g. an API call is needed but missing in the specs, hence not generated, the developer can add this functionality by augmenting the generated service.
A good example is `AugmentedPlansService`, where you fetch a `Plan` entity type (which has been generated) through the table interface: https://github.com/exense/step-frontend/blob/ac424bc3b8efe246f58993cbcc21e60140e70e69/projects/step-core/src/lib/client/augmented/services/augmented-plans-service.ts#L9 . In the generated `PlansService` there is no method to fetch plans, so the dev augments it.

 When the specs are updated and the client code is re-generated, the augmented code remains untouched. 

Another reason why one might want to augment the generated service is to patch/fix it, like in this method, where the client has to deal with non-JSON response. The method of the originally generated code fails (since the client fails because it expects JSON), so it is overwritten to make it work:
https://github.com/exense/step-frontend/blob/ac424bc3b8efe246f58993cbcc21e60140e70e69/projects/step-core/src/lib/client/augmented/services/augmented-scheduler-service.ts#L27

Yes, it brings duality, but as a rule of thumb it could work like that: import the generated service. If you can not find your desired method, look whether there exists an augmented version of it. If not, create your own augmentation.

projects/step-core/src/lib/client/augmented/services/augmented-executions-service.ts
dvladir: Yes let's discuss it

projects/step-core/src/lib/client/augmented/services/augmented-executions-service.ts
chillbirdo: Maybe then in a later API version, the endpoint that returned a non-JSON is fixed and returns a JSON. The client code could be regenerated and the augmented version of the service could be erased.
Same goes for the missing fetch-Plans call that has been added: Maybe a later API version includes this call so that it is not needed to query the table manually, so the augmented service could be erased.

So in my opinion it makes sense to augment the generated services because it is exactly we want: use them but maybe also extend/fix/modify them, but also only in some cases. Should we use one Augmented service for each generated one, even if there is no need to extend/fix/modifiy it, in order to access it only through this layer of abstraction? I think this is too much overhead and as I mentioned: using generated services directly is fine as long as one does not need extra functionality.

Do my arguments justify the occasional augmentation of the generated services for you?

projects/step-core/src/lib/client/augmented/services/augmented-executions-service.ts
chillbirdo: I was thinking about using the datasource like that:
```<step-table [dataSource]="_plansApiService.getPlansTableDataSource()"```, so not even having a reference of datasource in the component. But I agree, it is kind of useless, let us keep the datasource reference in the controller. Modified it in the latest commit

File: projects/step-core/src/lib/client/augmented/services/augmented-resources-service.ts [hunk 5]
```diff
@@ -0,0 +1,25 @@
...
+    name: 'attributes.name',
+    resourceType: 'resourceType',
+    id: 'id',
+  });
+
+  constructor(override httpRequest: BaseHttpRequest, private _tableRest: TableRestService) {
+    super(httpRequest);
+  }
+
+  public getResourcesTableDataSource(): TableRemoteDataSource<any> {
```
dvladir, line 22: @chillbirdo  For typescript the correct variant would be
```
private  _dataSource = new TableRemoteDataSource(...);

get dataSource(): TableRemoteDataSource {
  return this._dataSource;
}
```

So without set accessor we tell, that dataSource can not be redefined anywhere else.
But the shorter variant would be just have a public readonly field
```
readonly dataSource = new TableRemoteDataSource(...);
```

Probably it's another for discussion.
I understand that you follow the Java approach where field's assessors are made with getField/setField methods. I don't think that it make sense to follow the same approach for Typescript

dvladir (approved):

neogucky (approved):

PR #51, opened 2022-07-05
Sed 1105 refactor refactored tables fix spinner draft

chillbirdo:
This is a suggestion to remove boilerplate code like:
```
  private scheduledTask$ = this.scheduledTaskRequest$.pipe(
    tap((_) => ((this as InProgress).inProgress = true)),
    switchMap((_) => this._schedulerService.getScheduledExecutions()),
    tap((_) => ((this as InProgress).inProgress = false)),
    shareReplay(1)
  );
```
By the way, at the moment this code is useless, since the `inProgress` flag is solely here to show the spinner, which is never shown (i tested with delayed responses). In other words, **the spinner is broken. We might open a separate ticket for that.**

I put my suggestion in a draft branch (on top of [SED-1105-Refactor-refactored-tables](https://github.com/exense/step-frontend/tree/SED-1105-Refactor-refactored-tables) ) and I do not think myself it is the most elegant way to solve the problem (although it works and removes the boilerplate). It should most likely be solved inside the table or dataSource code.


dvladir:
@chillbirdo In case if you manually load the data and define the inProgress flag, its need to be passed to step-table as input parameter.
Is see that in original code of schedule-task-request.component.html inProgress parameter isn't set, probably that's why spinner doesn't work.

About TableAdapter. What do you think, if instead of service we will introduce a custom pipepable operator?

```
const switchMapWithProgress<REQ, RES>: (operation: (value: REQ) => Observable<RES>) => [BehaviorSubject<boolean>, Observable<RES>] = (operation) => (source$) => {
  const inProgress$ = new BehaviorSubject<boolean>(false);
  
  const result$ = source$.pipe(
    tap(_ => inProgress$.next(true)),
    switchMap(x => operation(x)),
    tap(_ => inProgress$.next(false)),
  );
  
  return [inProgress$, result$];
}
```

Then the invocation could be look like:
```
const [inProjgress$, scheduledTask$] = scheduledTaskRequest$.pipe(
switchMapWithProgress((_) => this._schedulerService.getScheduledExecutions())
);
```

chillbirdo:
Oops you are right **sorry for the confusion, the spinner is not boken.** It would not make any sense either, since it worked in my proposes solution, when the `inProgress` flag was set. :neckbeard:

Concerning your solution with the `switchMapWithProgress`-pipe: I would prefer to have the feature integrated in the table  without needing to manually pass the `inProgress` status. There could be a flag `showInProgress`, that is `true` by default, because normally we would want the spinner to be shown anyways right?

dvladir:
@chillbirdo 
For the remote data sources, the spinner works out of the box. In this case table automatically fetches data, when user changes the pages or invokes the sort.

For the local data sources the situation is different. Table isn't responsible for data loading. It doesn't know - was the data loaded form the back-end or was it formed on the front-end side as a simple array. 

From this point of view the responsibility to show the spinner belongs to table's parent component (or any other part the fetches the data), but it needs to notify table to display the spinner. That's why I add that additional input parameter that allows to control the spinner visibility.

Otherwise we can introduce a new implementation of TableDatasource, that will make a fetch on first page request and on reload's invocation (On pagination, sorting and search it will work like LocalPaginationDataSource). Probably it's initialization will look like:
```
readonly dataSource = new SingleFetchDataSource(() => this._schedulerService.getScheduledExecutions(), {
  searchPredicates: {
     ....
  },
  sortPredicates: {
    ...
  }
});
```
It will allow us to skip manual spinner control in most of scenarios.

chillbirdo:
if the outcome of that draft request is the creation of a new ticket for a `TableSource` implementation in the way suggested by @dvladir , I'd be perfectly happy. what do you think @neogucky ?

neogucky:
> if the outcome of that discussion is the creation of a new ticket for a `TableSource` implementation in the way suggested by @dvladir , I'd be perfectly happy. what do you think @neogucky ?

That sounds like the ideal solution. Can you create the ticket?

chillbirdo:
As discussed with @neogucky , I created the following ticket as a result of this discussion: https://exense.atlassian.net/browse/SED-1145

PR #55, opened 2022-07-11
SED-1064: Refactor Tables: functionPackageList.html

dvladir:


chillbirdo (approved):
Looks good to me! Thanks for applying the guidelines with service augmentation and separating out logic for opening modals

projects/step-frontend/src/lib/modules/function/components/function-package-link/function-package-link.component.ts
```diff
@@ -69,7 +67,7 @@ export class FunctionPackageLinkComponent {
   }
 
   loadFunctionPackage() {
-    this._httpClient.get<any>(`rest/functionpackages/${this.id!}`).subscribe((response) => {
+    this._api.get7(this.id!).subscribe((response) => {
```
neogucky, line 70: I don't like this function name, but I guess it is from our API-definition? Do we have more weird names there and should try to solve this?

neogucky (approved):
Looks good, but lets talk about get7 before merging

projects/step-frontend/src/lib/modules/function/components/function-package-link/function-package-link.component.ts
chillbirdo: I don't like it either, but that is the `operationId`, defined in the backend's OpenAPI specs. I assume that it is not set manually and therefore the the OpenAPI engine in the backend itself generates this name.
```
    "/functionpackages/{id}" : {
      "get" : {
        "tags" : [ "Keyword packages" ],
        "operationId" : "get_7",
        "parameters" : [ {
          "name" : "id",
          "in" : "path",
          "required" : true,
          "schema" : {
            "type" : "string"
          }
        } ],
```
(You can see yourself by navigating to https://stepos-master.exense.ch/rest/openapi.json and search for `get_7`)

The only sustainable solution is to refine the OpenAPI specs in the backend. The team who is building a client will always co-work on the specs in a back and forth manner, hence they are the only ones evaluating it...

chillbirdo: It could be added as another point to that list: https://trello.com/c/O5VoMq3h/75-commuinicate-openapi-spec-issue
I think our feedback of the specs is high in value. Having clean interfaces only improves the overall quality of the product.

chillbirdo: Most of the `operationId`s are set in a good way, e.g. 
```
    "/functionpackages/{id}/functions" : {
      "get" : {
        "tags" : [ "Keyword packages" ],
        "operationId" : "getPackageFunctions",
        "parameters" : [ {
          "name" : "id",
          "in" : "path",
          "required" : true,
          "schema" : {
            "type" : "string"
          }
        } ],
```
There are just a few that have weird names. It should be handled in the backend.

neogucky: Alright, I will add it to the tasks, we need to keep this in mind so we are not confused when these functions don't work anymore!

neogucky: And thanks for the detailed exlanation

PR #56, opened 2022-07-11
Sed 1052 refactor tables gridtokens.html

chillbirdo:


projects/step-frontend/src/lib/modules/grid/components/agent-list/agent-list.component.html
```diff
@@ -0,0 +1,58 @@
...
+  <ng-container matColumnDef="url" stepSearchCol>
+    <th mat-header-cell *matHeaderCellDef mat-sort-header>Url</th>
+    <td mat-cell *matCellDef="let element">
+      {{ element.agentRef!.agentUrl! }}
+    </td>
+  </ng-container>
+  <ng-container matColumnDef="type" stepSearchCol>
+    <th mat-header-cell *matHeaderCellDef mat-sort-header>Type</th>
+    <td mat-cell *matCellDef="let element">
+      <span class="agent-type" [class]="'agent-type-'.concat(element.agentRef!.agentType)"></span>
```
dvladir, line 16: Function call inside binding.
I would suggest to use additional pipe, that will form a required class string and then use it like:
```
<span [ngClass]="element.agentRef!.agentType | agentTypeClass"></span>
```

projects/step-frontend/src/lib/modules/grid/components/status-summary/status-summary.component.html
```diff
@@ -0,0 +1,37 @@
+<div
```
dvladir, line 1: This component looks very similar to `status-distribution.component`
Probably we can unify them or create a set of similar components in the same approach.

projects/step-frontend/src/lib/modules/grid/components/status-summary/status-summary.component.scss
```diff
@@ -0,0 +1,19 @@
+.token-status-IN_USE {
```
dvladir, line 1: As it was done for status-distribution, I think it will be good to put all colors in variables.scss and group them in map (See status-distribution.component.scss for example)

projects/step-frontend/src/lib/modules/grid/components/token-list/token-list.component.html
```diff
@@ -0,0 +1,103 @@
...
+  <ng-container matColumnDef="agent" stepSearchCol>
+    <th mat-header-cell *matHeaderCellDef mat-sort-header>Agent</th>
+    <td mat-cell *matCellDef="let element">
+      {{ element.agent!.agentUrl! }}
+    </td>
+  </ng-container>
+  <ng-container matColumnDef="attributes" stepSearchCol>
+    <th mat-header-cell *matHeaderCellDef>Attributes</th>
+    <td mat-cell *matCellDef="let element">
+      {{ _flatObjectFormatService.format(element.token!.attributes!) }}
```
dvladir, line 29: Code invocation in binding.
Introduce additional pipe, that invokes FlatObjectFormatService.
Or you can turn FlatObjectFormatService into pipe itslef in the following way
```
@Pipe({
  name:  'flatObjectFormat'
})
class FtabObjectFormatPipe implements PipeTransform {
  
  // static method to use it outside of templates
  static transform(obj: any): string {
       return JSON.stringify(obj)
                 .replace(/{/g, '')
                 .replace(/}/g, '')
                 .replace(/:/g, '=')
                 .replace(/,/g, ', ')
                 .replace(/"/g, '');
  }

  transform(obj: any): string {
     return FtabObjectFormatPipe.transform(obj);
  }

}
```

projects/step-frontend/src/lib/modules/grid/components/token-list/token-list.component.ts
```diff
@@ -0,0 +1,110 @@
...
+export class TokenListComponent implements OnDestroy {
+  readonly TYPE_LABEL_TRANSLATIONS: { [key: string]: string } = {
+    default: 'Java',
+    node: 'Node.js',
+    dotnet: '.NET',
+  };
+
+  readonly inProgress: boolean = false;
+
+  readonly json = JSON;
```
dvladir, line 41: What is it used for?

dvladir (changes requested):

projects/step-frontend/src/lib/modules/grid/components/token-list/token-list.component.html
chillbirdo: thanks for the suggestion with the static method, since I need it in the controller as well

projects/step-frontend/src/lib/modules/grid/components/token-list/token-list.component.html
```diff
@@ -0,0 +1,103 @@
...
+      </div>
+    </td>
+  </ng-container>
+
+  <ng-container matColumnDef="actions">
+    <th mat-header-cell *matHeaderCellDef>Actions</th>
+    <td mat-cell *matCellDef="let element" class="cell-actions">
+      <step-btn-group>
+        <button
+          *ngIf="element.state! != 'MAINTENANCE'"
```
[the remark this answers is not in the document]
chillbirdo: I know but sometimes forget, thanks for the reminder

projects/step-frontend/src/lib/modules/grid/components/token-list/token-list.component.ts
chillbirdo: nichego takogo!

projects/step-frontend/src/lib/modules/grid/components/token-list/token-list.component.html
chillbirdo: I not going to use methods in the template any more, but more (pure) pipes

chillbirdo:
new PR unites 1052 and 1053: https://github.com/exense/step-frontend/pull/65

PR #61, opened 2022-07-13
Sed 1048 refactor tables plan selection table

chillbirdo:


dvladir (approved):

neogucky (approved):
Sounds good, we can't test it because of the DNS issue so I will do a tentative merge and revert in worst case

PR #62, opened 2022-07-13
SED-1129: Simplify enterprise plugin structure

dvladir:


projects/step-frontend/src/lib/plugins-initializer/plugins-initializer.ts
```diff
@@ -3,14 +3,18 @@ import { LegacyPluginDefinition } from './shared/legacy-plugin-definition';
...
+// Ignore original multitenancy since it's become part of enterprise core
+const IGNORE_PLUGINS: ReadonlyArray<string> = ['multitenancy'];
 
 // For testing purposes only
 // Allows to add plugins, that don't returned from BE
-const ADDITIONAL_PLUGINS: ReadonlyArray<string> = [
-  // Add strings like this 'pluginName/remoteEntry.js'
+const ADDITIONAL_PLUGINS: ReadonlyArray<MicrofrontendPluginDefinition> = [
+  // Add object like this {name: 'pluginName', entryPoint: 'pluginName/remoteEntry.js' }
+  { name: 'stepEnterpriseCore', entryPoint: 'step-enterprise-core/remoteEntry.js' },
```
neogucky, line 17: @dvladir  So I guess we want to send this from the BE, right? Also, we need a way to manage which plugins are loaded. It is also ok, to use a flag to check if the BE has these plugins activated. Did you think about this?

projects/step-frontend/src/lib/plugins-initializer/plugins-initializer.ts
```diff
@@ -24,34 +28,32 @@ const fetchDefinitions = async (): Promise<PluginDefinition[]> => {
...
-const loadPlugins = (compiler: Compiler, injector: Injector) => {
+const loadPlugins = (compiler: Compiler, injector: Injector, registry: PluginInfoRegistryService) => {
   return async () => {
     const pluginDefinitions = await fetchDefinitions();
     if (pluginDefinitions.length === 0) {
       return;
     }
 
-    const { legacy, microfrontend } = pluginDefinitions.reduce(
+    let { legacy, microfrontend } = pluginDefinitions.reduce(
```
neogucky, line 56: I know microfrontend is not a new variable/class but it is a bit dificult to understand for me. What exactly is the name standing for?

projects/step-frontend/src/lib/plugins-initializer/plugins-initializer.ts
```diff
@@ -66,6 +68,19 @@ const loadPlugins = (compiler: Compiler, injector: Injector) => {
...
+      ...microfrontend.map((pluginInfo) => pluginInfo.name),
+    ];
+    registry.register(...pluginNames);
+
+    legacy = legacy.filter((l) => {
+      const ignoredPlugins = IGNORE_PLUGINS.filter((name) => l.angularModules.includes(name));
+      return ignoredPlugins.length === 0;
+    });
+
+    microfrontend = microfrontend.filter((m) => !IGNORE_PLUGINS.includes(m.name));
```
neogucky, line 82: I'm not entirely sure what happens here. Is this the part that solves my question earlier about how to manage which plugins to enable?

If yes, can we change the code so that it uses ENABLED_PLUGINS instead of IGNORED_PLUGINS so it is more clear what happens here?

neogucky (changes requested):
Not really changes requested but it is depending on ypur answers ;)

projects/step-frontend/src/lib/plugins-initializer/plugins-initializer.ts
dvladir: @neogucky Right now stepEnterprizeCore analyzes the response from '/rest/app/plugins/' to determine, should it include one of it parts or not.
Probably in future, it will be better to use additional endpoint to determine it.

Yes, the link to stepEnterpriseCore, should be also added to that rest endpoint response

projects/step-frontend/src/lib/plugins-initializer/plugins-initializer.ts
dvladir: We have to types LegacyPluginDefinition and MicrofrontendPluginDefinition
We receive an array pluginDefenitions, that may contain elements of both of these types.
Then the reduce is called, and aggregates items of each type to corresponding array - `legacy` for LegacyPluginDefinition items and `microfrontend ` for MicrofrontendPluginDefinition items

projects/step-frontend/src/lib/plugins-initializer/plugins-initializer.ts
dvladir: I don't think that such names will describe exactly what's happened here.
In lines 71 - 75, we collect the names of **all** plugin's modules that we were able to receive. Then we are putting it to PluginRegistry service.
PluginRegistry is just a simple storage - later StepEnterpriseCore will use it to determine, which part should be enabled.

Then as backend returns some old data (like link to multienancy), we need to exclude it, to prevent it's initialization as it has become a part of StepEnterpriseCore. That was the INGNORE_PLUGINS do.
In future as all plugins will be moved to StepEnterpriseCore and rest endpoint will be updated, INGNORE_PLUGINS should be removed from the code

neogucky (approved):

PR #63, opened 2022-07-15
SED-1140 Created new login page

neogucky:


dvladir (approved):

chillbirdo (approved):

chillbirdo:
supplement: the only thing that I am missing is that you can not just press Enter after entering the password as you could in the old version..

PR #64, opened 2022-07-18
SED-1057: refactor tables execution step

dvladir:
This task intersects much with SED-1056, because it uses the same controller in angularJS.
To implement SED-1057, I had to migrate the panels management & table's entity selection. Then I downgraded that's part and integrated in the angularjs `executionProgress` directive's controller.
To make it clear, what was changed in the original code, I've just commented the old functionality.

Keyword Call table has additional task, that's why I just wrapped it in the upgraded component and insert in the new layout as is


chillbirdo (approved):

neogucky (approved):
Nice work, I like the generics

PR #65, opened 2022-07-18
Sed 1053 sed 1052 refactor grid

chillbirdo:


projects/step-frontend/src/lib/angularjs/partials/grid/gridTokenGroups.html
```diff
@@ -1,38 +1,3 @@
...
-                    </st-column>
-                </st-columns>
-            </st-table>
-        </div>
-    </div>
-</div>
\ No newline at end of file
+<div>
+  <step-token-group-list></step-token-group-list>
+</div>
```
neogucky, line 3: I noticed that in most cases we don't need the old surrounding container. These where previously used to mark the controller but at least for all views I migrated I could remove the div without any problems.

projects/step-frontend/src/lib/modules/_common/components/execution-link/execution-link.component.ts
```diff
@@ -0,0 +1,22 @@
...
+  @Input() executionId?: string;
+  @Input() executionDescription?: string;
+
+  constructor(private _executionService: ExecutionsService) {}
+
+  ngOnChanges(): void {
+    if (this.executionId && !this.executionDescription) {
+      this._executionService.get3(this.executionId!).subscribe((ex: Execution) => {
+        this.executionDescription = ex?.description;
+      });
```
neogucky, line 19: This might be executed unexpectedly and request resources from the BE. 

You should prbably check if changes.executionId.currentValue !== changes.executionId.previousValue or changes.executionId.executionDescription!== changes.executionId.executionDescription before doing the request.

see:
[ngOnChanges(changes: SimpleChanges): void](https://angular.io/api/core/OnChanges#ngOnChanges)

projects/step-frontend/src/lib/modules/grid/components/token-list/token-list.component.ts
```diff
@@ -0,0 +1,85 @@
...
+      type: (element, searchValue) =>
+        TokenTypeComponent.TYPE_LABEL_TRANSLATIONS[element.token!.attributes!['$agenttype']]!.toLowerCase().includes(
+          searchValue.toLowerCase()
+        ),
+      agent: (element, searchValue) => element.agent!.agentUrl!.toLowerCase().includes(searchValue.toLowerCase()),
+      attributes: (element, searchValue) =>
+        FlatObjectStringFormatPipe.format(element.token!.attributes!).includes(searchValue.toLowerCase()),
+      state: (element, searchValue) => searchValue.toUpperCase().includes(element.state!.toUpperCase()),
+      // prettier-ignore
+      //@ts-ignore
```
neogucky, line 38: Why do we use so many ts-ignore in this file? I looked at the errors and it seems we could just add the missing properties to the models or am I missing something?

neogucky (changes requested):
Looks pretty good already, I added some smaller points though. 

Also there seems to be a build error, probably in how the new Pipe is registered. (locally this usually works since in the workspace environment core and frontend are seens as one) https://build.exense.ch/#/root/executions/62d5d89187a5284560d72f8a?tenant=build

projects/step-frontend/src/lib/modules/_common/components/execution-link/execution-link.component.ts
chillbirdo: see https://github.com/exense/step-frontend/pull/59#discussion_r924244647

projects/step-frontend/src/lib/modules/grid/components/token-list/token-list.component.ts
chillbirdo: introduced a new type that has all the fields that are returned by the API, so got rid of ts-ignore :+1:

chillbirdo:
implemented all of your feedback, can be merged imo, https://github.com/exense/step-frontend/pull/59 and https://github.com/exense/step-frontend/pull/56 closed :+1:

neogucky (approved):
Looks good now!

PR #66, opened 2022-07-19
SED-1029 new timeseries

neogucky:


lucianbaciu:
Features:

- Implement uPlot chart library
- Add different type of charts
- Syncronized charts
- Time-ranger
- Datepicker similar to Grafana implementation
- Loading effects
- Realtime execution capability (auto refresh, refresh interval, etc)

TODO list:
- [ ] Remove all @ts-ignore and migrate all native js to Typescript
- [ ] Try to split ExecutionPageComponent into multiple smaller components
- [ ] Complete filtering/grouping logic
- [ ] Handle corner cases like no data, etc
- [ ] Ability to switch throughput (TPH/TPS/TPM)
- [ ] Complete table features (serch/sort/design)

projects/step-frontend/src/lib/modules/timeseries/chart/time-series-chart.component.scss
```diff
@@ -4,3 +4,60 @@
   max-height: 100%;
   padding-top: 24px;
 }
+
```
dvladir, line 7: Style encapsulation turned off for this components.
This style may intersect with some other styles in applications.
If `ViewEncapsulation.None` is required for current case, it make sense add parent selector for each rule, to prevent intersection. The simplest way would be just to wrap styles as following:
```
step-timeserises-chart {
    
   .chart-wrapper {
             ....
   }
  
  .ranger {
    ...
  }

  .ts-tooltip {
    ...
  }

 ...
}
```

projects/step-frontend/src/lib/modules/timeseries/chart/time-series-chart.component.ts
```diff
@@ -10,17 +10,23 @@ import {
...
 } from '@angular/core';
 import { TSChartSeries, TSChartSettings } from './model/ts-chart-settings';
 import { UplotSyncService } from './uplot-sync-service';
 import { UPlotUtils } from '../uplot/uPlot.utils';
+import { tooltipPlugin } from './tooltip-plugin';
+import { ExecutionTabContext } from '../execution-page/execution-tab-context';
+import { TSTimeRange } from './model/ts-time-range';
+import { AlignedData } from 'uplot';
 
 declare const uPlot: any;
```
dvladir, line 23: I've tried to play with it to avoid any, and found the following example:
https://stackblitz.com/edit/angular-ivy-k6ig7h?file=src%2Fapp%2Fapp.component.ts

In step, I was able to include it in the following way
```
import { AlignedData, Options, Cursor } from 'uplot'; // required uplot types
//@ts-ignore
import uPlot = require('uplot'); // uplot class itself
```
Yes we use `@ts-ingore` to write such import, but it forces the IDE to check uPlot types more strict and as a result write better code. `declate const uPlot: any` wont' require any more

Interesting moment, I've checked the assembled scripts and didn't notice that this line add any real import. So we still need to include uPlot sources in angular.json scripts sections.
But in the example from the stackblitz, it looks like it work as a real import. That's the point for further investigations

neogucky: We should do the change now @lucianbaciu and create a ticket for solving the import vs. angular.js issue. (Would be great not to have uplot defined in there)

projects/step-frontend/src/lib/modules/timeseries/chart/time-series-chart.component.ts
```diff
@@ -46,7 +55,14 @@ export class TimeSeriesChartComponent implements OnInit, AfterViewInit, OnChange
...
-  ngAfterViewInit(): void {
+  resetZoom(): void {
+    this.uplot.setData(this.uplot.data, true);
+  }
+
+  /**
+   * @param settings
+   */
+  createChart(settings: TSChartSettings): void {
     let getSize = () => {
```
dvladir, line 66: We will have to check the performance in production cases.
Right now chart rendering works inside of angular's change detection cycle.
Probably it make sense to move it's invocation in zone.runOutsideAngular() to split charts rendering from angular's underhood stuff.

neogucky:
@lucianbaciu why has the [package-lock.json](https://github.com/exense/step-frontend/pull/66/files#diff-053150b640a7ce75eff69d1a22cae7f0f94ad64ce9a855db544dda0929316519) changed? Since I see no changes to package.json it should not have been re-generated.

chillbirdo:
First of all **thanks for your work @lucianbaciu , metrics look good and works fine** for most cases!

I had a bit of a hard time reviewing it, since short time fixed could never fix the underlying architectural problems that exist here. I gave reviews from three different perspectives:

## as User:
- remove the (X) link, which leading to an errorneous page
- when the execution has a `TECHNICAL_ERROR`, the metric page hangs. I opened a ticket: https://exense.atlassian.net/browse/SED-1192
- "Average Response Time" shows same time on each point in time (990 in this example: `[server]/#/root/executions/62e7f6cd2ab08142ebbc9f9e`)
- add help hints: what do the charts say? e.g. "Average Response Time" of what? It is not self explanatory.

## as Coder:
- commented-out code = code smell
	- half of `uplotOptions.ts`
	- `tooltip-plugin.ts:234`
- renaming and documentation: eg.`tooltip-plugin.ts:22` who would ever want to touch this function? please rename variables and document what it happening there OR: refer where it is pasted from in a comment or separate readme. Can the pasted code be interchanged with a new version of the lib?
- `time-series-service.ts` vs generated `TimeSeriesService.ts` + Req-Resp-DTOs: generated could/should be used


## as SW Engineer / Architect:

First of all I have to say that I do not know the exact requirements of the Metrics and I did not do a research whether or not there would have been more out-of-the-box solutions that could have been configured to meet the requirements. I assume that this has been done extensively and the result was that there was no way around doing low-level calculations manually (as done for example in `tooltip-plugin.ts`), instead of using a clean, well documented configuration interface of an existing, Angular compatible package.

The biggest problem I see with the actual solution is the **maintainability**. I believe that nobody will ever want to touch this code in the future, no matter how many `@ts-ignore` we can eliminate. In the case that a change will be required in the future and Lucian would not be available, architects will consider other solutions than letting someone dig into this code, simply because of scarcity of time.

The solution to this problem would be IMO:
- move this whole code to a separate project and import it as a package/plugin in STEP
- create a configurable interface for users of this package/plugin.
- write a documentation for the interface for how to use it (just like the docs of NPM packages)
- write comments to your code, especially for the low level calculation stuff

I don't know how much value I am contributing with my analysis from the SW engineer perspective, because I assume we do not have enough resources for that. Nevertheless, communication of idead is never wrong, and it probably helps for the decision making in the future.

chillbirdo:
@lucianbaciu I can just imagine how hard it was to put this all together and make it work! :muscle: I think your work will make customers happy! My critiques are not meant to frustrate you, but to exchange ideas on the solution! :v:

neogucky:
I will try not to mention anything mentioned before:

**Structuring / Naming**
- execution-page.component seems to be the performance view and not the execution-(details)-view. 
- ...partials/execution/executionViz.html should be renamed executionPerformance (This seems to be an old leftover and the reason for your naming)
- Probably in the timeseries menu we should structure the code so that it is clear we have some general chart components  and probably two views to show them: executionPerformanceView (what you currently call execution-page) and taskPerformanceView (what you see when clicking on dashboard in scheduler view)
- I saw a lot of models not in the models folder and did not see if there is a reason

**General code style**
I agree with Gilbert that the minimalistic code in tooltip-plugin isn't nice. We should definetely try to make them humn understandable by nice naming and probably splitting it into multiple helpfully named functions.

I also see a lot of improvement to how it was before, so it is going in the right direction! I would definetiely prefer to have another month to do the fixes mentioned here but let's start with creating tickets and continue from there. @lucianbaciu can you create tickets and maybe comment on the code reviews to either clarify the issue or link to the ticket?

chillbirdo:
Thinking structure, I wished that it would be easy to recognize what code belongs to which of the following parts:

_(**A** Data Receiving) -> (**B** Data Transformation) -> (**C** Plot drawing)_

ad A: no new code required. Generated Services an DTOs should be used
ad B: encapsulate and add documentation: how does the data from the server need to be transformed to feed C?
ad C: encapsulate and add documentation: can 3rd party libs be updated and how? did you extend them? why and in what way?

chillbirdo (dismissed):
see comments above

projects/step-frontend/src/lib/modules/timeseries/chart/time-series-chart.component.scss
lucianbaciu: will fix this!

projects/step-frontend/src/lib/modules/timeseries/chart/time-series-chart.component.ts
lucianbaciu: will fix this!

projects/step-frontend/src/lib/modules/timeseries/chart/time-series-chart.component.ts
lucianbaciu: this was initially my plan. @neogucky told me to do it without, and think about it just if we have performance issues. currently it works good. I will create a task for it so we don't forget

lucianbaciu:
guys, no worries about frustrations or other things like that. I'm aware about most of the things that you specified, and I agree with them. I'm gonna fix what can be easy, and create tickets for the other. I will try to focus on the documentation/comments too. It is a big thing with a lot of research behind, but is nowhere documented.
Thank you!

neogucky:
@lucianbaciu any update on the open points? Did you fix anything essential and create tickets for the rest so we can do the merge?

PR #68, opened 2022-07-19
SED-1143: Dynamic table fields

dvladir:
`step-custom-columns` component was introduced. It works nearly the same as original st-custom-columns - loads information about additional columns from backend and columns definitions base on this information.

For now we don't support custom HTML template, the display value is always taken by path specified in the ID property.
As we previously added some columns directly, there are columns duplicates in some tables


neogucky (approved):

neogucky:
@chillbirdo I already merged since this is a blocking PR, but still take a look at the changes (and if you find problems we can also solve them with another PR)

projects/step-core/src/lib/modules/table/components/custom-columns/custom-columns.component.ts
```diff
@@ -0,0 +1,88 @@
...
+  columns: ColInput[] = [];
+
+  readonly trackByCol: TrackByFunction<ColInput> = (index, item) => item.id;
+
+  @ViewChildren(MatColumnDef) colDef?: QueryList<MatColumnDef>;
+  @ViewChildren(SearchColDirective) searchColDef?: QueryList<SearchColDirective>;
+
+  constructor(private _screens: AugmentedScreenService) {}
+
+  ngOnChanges(changes: SimpleChanges): void {
```
chillbirdo, line 39: Why are you using `ngOnChanges` here? The way this component is used, `this.screen` and `this.excludeFields` are constants and it would not make any sense to ever use variables here? `ngOnInit` would be sufficient:
```
  ngOnInit(): void {
      this.loadColumns(this.screen, this.excludeFields);
  }
```

projects/step-core/src/lib/modules/table/components/custom-columns/custom-columns.component.ts
```diff
@@ -0,0 +1,88 @@
...
+  @Output() columnsReady = new EventEmitter<unknown>();
+
+  columns: ColInput[] = [];
+
+  readonly trackByCol: TrackByFunction<ColInput> = (index, item) => item.id;
+
+  @ViewChildren(MatColumnDef) colDef?: QueryList<MatColumnDef>;
+  @ViewChildren(SearchColDirective) searchColDef?: QueryList<SearchColDirective>;
+
+  constructor(private _screens: AugmentedScreenService) {}
```
chillbirdo, line 37: the name `_screens` lets me think of an array of objects or primitives. in this case `screen` even is a primitive. For the injected service, a name like `_screensApiService` would make the code more readable imo

projects/step-core/src/lib/modules/table/components/custom-columns/custom-columns.component.ts
dvladir: Since these fields are inputs, they can be changed at runtime (event we don't have such cases now).
So I just wish to support the ability to reconfigure custom columns, in case if it's parameters will be changed.

chillbirdo: Technically I understand, but in this case I can not see a usage for this component having the input changed. But right, maybe there will be a usage in the future, so why not supporting it.. :+1:

chillbirdo (approved):

PR #74, opened 2022-07-26
SED-1143: Rework tables API (FE)

dvladir:


projects/step-core/src/lib/modules/table/shared/table-remote-data-source.ts
```diff
@@ -41,43 +44,31 @@ export class TableRequest {
...
+        direction: req.orderBy.order === 'asc' ? SortDirection.ASCENDING : SortDirection.DESCENDING,
+      }
+    : {
+        field: req.columns[0],
+        direction: SortDirection.ASCENDING,
+      };
+
+  // if (req.filter) {
+  // todo vad: for now it is not clear how to proceed queries like filter: "not(resourceType=attachment)'"
+  // }
```
neogucky, line 71: @dvladir this can no be done like this, I had not had time to update the code, can you quickly add this?

 {"oql":"not(resourceType=attachment)"}

neogucky (changes requested):
Looks good to me! I added the open point but it is defenitely a nice improvement

chillbirdo (dismissed):
Looks good to me. I only rechecked it now, since I reviewed it once before I went to vacation but did not approve because it was/is still a draft.

neogucky (approved):
Looks good to me and also the oql filters looked correct in https://stepos-sed-987-new-table-api.exense.ch/#/root/resources

neogucky (approved):

PR #76, opened 2022-08-01
SED-1141: Create Registry concept for NG2+

dvladir:


neogucky (approved):
I like it. Having to create multiple wrapper components in order to make different types work is a bit of a hassle but I don't see a cleaner solution.

@dvladir are we ready to move everything to the new registry after this is merged in?

chillbirdo:
I would wish to have one working example of something registered in the old way and now working in the new way. :pray:

projects/step-core/src/lib/modules/custom-registeries/services/base-registry.service.ts
```diff
@@ -0,0 +1,33 @@
+import { CustomRegistryType } from '../shared/custom-registry-type.enum';
+import { CustomRegistryService } from './custom-registry.service';
+import { Type } from '@angular/core';
+import { CustomRegistryItem } from '../shared/custom-registry-item';
+
+type ItemInfo = Pick<CustomRegistryItem, 'type' | 'label'>;
+
+const convert = (item?: CustomRegistryItem): ItemInfo | undefined => {
```
chillbirdo, line 8: a general question: what is the advantage of defining this piece of code outside of the Service? Why not define it as a private method?

chillbirdo (dismissed):

projects/step-core/src/lib/modules/custom-registeries/services/base-registry.service.ts
dvladir: In current case `covert` is a very simple pure function that extract type and label, and it doesn't manipulate with service's data. So in my understanding, it doesn't belong to service.

In case if it would meet once in the service's code, it would be inline arrow function, but as it is appeared twice it was moved separately to prevent the code duplication.

dvladir:
@chillbirdo 
> I would wish to have one working example of something registered in the old way and now working in the new way. 🙏

Old usage example:
Registering
```
  .run(function (FunctionTypeRegistry, ViewRegistry, EntityRegistry) {
    ....
    EntityRegistry.registerEntity(
      'KeywordPackage',
      'functionPackage',
      'functionPackage',
      '/rest/functionpackages/',
      '/rest/functionpackages/',
      'st-table',
      'functionpackages/partials/functionPackageSelectionTable.html'
    );
  })
```

Usage
```
// From selectEntityOfType.html
<select-entity type="type" multiple-selection="multipleSelection" on-selection="select" handle="selectEntityHandle" />

// selectEntity source
   .directive('selectEntity', function () {
    return {
      restrict: 'E',
      scope: {
        type: '=',
        multipleSelection: '=?',
        onSelection: '=?',
        handle: '=?',
      },
      template: '<ng-include src="templateUrl" />',
      controller: function ($scope, EntityRegistry) {
        var entityType = EntityRegistry.getEntityByName($scope.type);

        $scope.tableHandle = {};

        $scope.templateUrl = entityType.templateUrl;

        $scope.notifySelection = function (selection) {
          if ($scope.onSelection) {
            $scope.onSelection(selection);
          }
        };
 ....
}
```
So Entity Registry links the entity name with template. And then selectEntity directive get this template by name and pass it to ng-include.

For the modern version the same will look like:
So instead of template link  'functionpackages/partials/functionPackageSelectionTable.html' we will have a component class. Let's call it like FunctionPackageSelectionTableComponent. So the registration would be like:
```
// We pass the direct reference to component class
// The register method has less parameters than an original one. 
// I discussed it with Tim, we decided to drop them, as they are not used.
entitryRegistry.register('functionPackage', 'KeywordPackage', FunctionPackageSelectionTableComponent);
```
And the usage will be like this:
```
<step-entity entity="functionPackage"></step-entity>
```
step-entity component will retrieve the information and render the component, that was associated with name "functionPackage"

dvladir:
> I like it. Having to create multiple wrapper components in order to make different types work is a bit of a hassle but I don't see a cleaner solution.
> 
> @dvladir are we ready to move everything to the new registry after this is merged in?

@neogucky We will have to migrate each entities templates one by one, and put it into this new registries.
Until it will be done, it's better not to remove the old registries and continue to use it.
I think that this new one is required to provide custom columns templates for the migrated tables, first of all.

PR #77, opened 2022-08-02
SED-1161 infinite loop grafana

neogucky:
Fixes https://exense.atlassian.net/browse/SED-1161

Also contains pipe that is not currently used but will replace function call when migrating to NG2+

### DIFF
--- a/projects/step-core/src/lib/services/dashboard.service.ts
+++ b/projects/step-core/src/lib/services/dashboard.service.ts
@@ -42,13 +42,11 @@ export class DashboardService {
   }
 
   private makeDashboardLink(taskId: string, isGrafanaAvailable?: boolean) {
-    let result: string;
     if (isGrafanaAvailable) {
-      result = '/#/root/grafana?d=3JB9j357k&orgId=1&var-taskId_current=' + taskId;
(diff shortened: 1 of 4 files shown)


lucianbaciu (dismissed):

projects/step-core/src/lib/services/dashboard.service.ts
```diff
@@ -4,42 +4,69 @@ import { AuthService } from './auth.service';
...
+  private makeDashboardLink(taskId: string, isGrafanaAvailable?: boolean) {
+    if (isGrafanaAvailable) {
       return '/#/root/grafana?d=3JB9j357k&orgId=1&var-taskId_current=' + taskId;
     } else {
       return '/#/root/dashboards/__pp__RTMDashboard?__filter1__=text,taskId,' + taskId;
     }
   }
 
-  checkAvailability(override = false) {
+  checkAvailability(taskId: string) {
```
lucianbaciu, line 52: Since it is a void method, I would add a /** documentation on top of the method, to explain what it does behind. Will it raise an error, will it set some flags? etc..

neogucky: This method is only for the class itself so no documentation is needed. But you are right, we should mark it as private in this case.

PR #78, opened 2022-08-08
SED-1169 regenerated client code from updated specs

chillbirdo:


chillbirdo:
turned out that the two files are exactly identical at the moment:
```>> curl https://stepos-nhy.exense.ch/rest/openapi.json | md5sum
474bc67187a9176c8bafb32c4b656455  -
```
```>> curl https://stepos-nhy.exense.ch/rest/private-openapi.json | md5sum
474bc67187a9176c8bafb32c4b656455  -
```

projects/step-core/src/lib/client/augmented/services/augmented-grid-service.ts
```diff
@@ -0,0 +1,36 @@
+import { Injectable } from '@angular/core';
```
chillbirdo, line 1: @neogucky this is the workaround for the removed methods from the openapi.json

chillbirdo: and btw: this is the workaround for the method that is defined as returning `application/json` but returning plain text: https://github.com/exense/step-frontend/blob/8c27bb0877f2b95128ad87aabae9860e8edfa7c8/projects/step-core/src/lib/client/augmented/services/augmented-scheduler-service.ts#L27

neogucky: Sounds good. I would imagine that the fix for grid will probably be not what you expect but another openapi file or something like that.

neogucky:
> 

Can confirm for both os and ee, will check in with BE.

neogucky:
@chillbirdo I checked with BE and the public/private api issue is technical and will stay for now. What happens is, the first API that is generated will be shown for both requests. So you need to request the private specs first, best on the master for now so the other link hasn't been called yet. Can you recheck if the generated client has the full features of the private specs?

dvladir:
@chillbirdo, @neogucky 
I've updated the PR.
Duplicated interfaces were removed. I decided to keep the table service and some table types, that I had written before - because it allows to specify the table response's items by generic type. Generated service doesn't allow this. But I changed the implementation to invoke the generated service under the hood, instead of forming the request manually.

chillbirdo:
@dvladir @neogucky Looks great, can be merged IMO!

projects/step-core/src/lib/client/generated/services/EntitiesService.ts
```diff
@@ -0,0 +1,96 @@
...
+        });
+    }
+
+    /**
+     * @param type
+     * @param requestBody
+     * @returns any default response
+     * @throws ApiError
+     */
+    public save1(
```
dvladir, line 81: Such methods with a number at the end of the name, still remain. Should we remove it, or will it be done later?

neogucky: Are there still methods in the API?

As an explanation, the numbers are added by swagger when a method name is used multiple times (i.e. delete for /plan/ and for /keaywords/) but David tried to remove the functions and I did a quick check and it looked good. If there are still numbers in the API we should ask him to rename the functions.

chillbirdo: when opening https://stepos-master.exense.ch/rest/private-openapi.json
and searching for `_1` I can still find 4 occurences of such method names, one of them beeing `save_1`, resulting in the generated code you mentioned above. So the generator did all good, the flaw is still in the API-specs.. but there are only 4 remaining :dancers: :dancers:

chillbirdo: I suggest to merge the branch. Sooner or later we will have to regenerate again anyways due to API changes

PR #79, opened 2022-08-09
SED-1170 introduces new npm script for local build

chillbirdo:
when developing inside the workspace locally and one needs to build the frontend project (for example to check on TS errors), they need to run `npm run build:local` from now on, otherwise it will fail due to the lack of `uplot` dependencies inside the `node_modules` directory. When working inside the workspace, this dependency is inside the `node_modules` of the workspace.

some remarks:
`uplot` is the only library that is loaded globally so far, that is why this case never occured so far. A quick google search did not reveal any uplot+angular integration examples :/, but I guess there should be a way to use it through `import` as well. We could take some time to solve this in another ticket if demanded.


chillbirdo:
ticket: https://exense.atlassian.net/browse/SED-1170

neogucky (approved):
Looks good, thank you.

PR #80, opened 2022-08-10
SED-1055: Refactor Tables. keywordsCalls.html

dvladir:


projects/step-core/src/lib/modules/table/components/table/table.component.ts
```diff
@@ -242,4 +255,12 @@ export class TableComponent<T> implements AfterViewInit, OnChanges, OnDestroy, T
       this.setupDatasource(cDatasource.currentValue);
     }
   }
+
+  exportAsCSV(): void {
+    if (!this.tableDataSource) {
+      console.log('No datasource for export');
```
neogucky, line 261: In general I would say hiding errors in the console is not best practise, but I think this should normally never happen so it is fine. But please use `console.error` for these cases.

projects/step-core/src/lib/modules/table/shared/table-local-data-source.ts
```diff
@@ -176,4 +176,8 @@ export class TableLocalDataSource<T> implements TableDataSource<T> {
   reload(): void {
     this._request$.next(this._request$.value);
   }
+
+  exportAsCSV(params?: unknown): void {
+    console.log(`TableLocalDataSource doesn't support export to CSV`);
```
neogucky, line 181: `    console.error('TableLocalDataSource doesn't support export to CSV');`

projects/step-core/src/lib/modules/table/shared/table-remote-data-source.ts
```diff
@@ -36,6 +37,7 @@ export class TableRequest {
     this.start = data?.start || undefined;
     this.length = data?.length || undefined;
     this.filter = data?.filter || undefined;
+    this.params = data?.params || undefined;
```
neogucky, line 40: I know it is not new, but isnt the undefined redundant? If either data or params is not defined it will be undefined.

neogucky (dismissed):
Great PR. I made some comments that should be fixed but they are more cosmetic.

projects/step-core/src/lib/modules/table/shared/table-remote-data-source.ts
dvladir: It's just a cleanup in case if parameter will be send as null, or empty string.
For example without it, we could receive the following case in theory, that request will contain:
```
{
  "filter": "",
  "params": null
}
```
And it may affect on the response's result.

With undefined replace, such fields won't be included in the requests json body.

neogucky (approved):

PR #83, opened 2022-08-12
SED-1166-use-new-custom-component

dvladir:


neogucky (approved):
Looks good! We still have to wait for the BE changes (currently in progress)

After BE changes are ready we have to check that all currently used valueHTMLTemplates need to be replaced with customComponents (part of BE but we have to check if we need new componnts for this)

neogucky (approved):
Looks good!

PR #84, opened 2022-08-15
SED-1111-table-interface-design

dvladir:
entities-selection module, doesn't know anything about tables or bulk operations. It just provides functionality to select entities. The module contains following parts:

- **selection-collector** - an object that stores information about the selection
- **entity-selection** - the check-box that links a single entity with selection-collector. When check-box is checked it means that **selection-collecoter** has information, that entity is selected
- **bulk-selection** - the check-box for bulk selection

There are two ways how selection-collector can be created and passed to the entity-selection and bulk-selection
- Create **selection-collector** by the **selection-collector-factory.service**, pass created instance as input parameter to **selection-collector** and **bulk-selection**. This approach might be useful, when multiple selections will required in scope of single component.
- Through the DI. Register **selection-collector** as component's provider by **selectionCollectionProvider** function. Use **entity-selection-di** and **bulk-selection-di** components. This approach is simpler, because there is no need to think about passing **selection-collector**. But for some cases it might be not so flexible.

Keeping this two approach might be complicated. Probably it make sense to keep only one of them in future.

Table's **bulk-operations** is implemented over the **bulk-selection**, it's add functionality of invoking specific operations and retrieves filter information from the table. **bulk-operations** uses the DI approach for bulk selection.


chillbirdo (approved):
No complains, good job

chillbirdo:
If we keep both ways of using, I would recommend to document it somewhere, probably in the component's code. Otherwise, the knowledge will get lost inside a closed PR..

neogucky (commented):
This looks very good, I'm just waiting with approval until the missing bulk selections are added

chillbirdo (dismissed):

neogucky (approved):
Great!

PR #85, opened 2022-08-16
SED-987: Adapt legacy tables to new API

dvladir:


neogucky (approved):
Looks great and simple! We are doing a smoke test and if all tables are working we can merge it in.

neogucky (approved):

PR #87, opened 2022-08-17
Sed 1174 layout redesign

chillbirdo:
# the new Layout is here!!! :star_struck: 

feel free to try it out!

EE Branch: https://bitbucket.org/exenseteam/step-enterprise-frontend/branch/SED-1174-layout-redesign

## features:
- CSS variables are used to easily readjust alignments
- menu is build dynamically from ViewRegistry

## todos:

The following is still to be done. Feel free to extend/edit this list.

### visually:
- [x] add all icons (done with temporary icons - please ask the designer for Material Icon infos)
- [x] change collapsible-sign `^` to appropriate icon
- [x] highlight the currently active menu point (can be done in pure CSS)
- [x] round the bottom-most edges of the collapsible menu
- [x] deal with an uncollapsed menu that is larger in height than the screen
- [x] login dialog is misaligned
- [x] remove upgrade box in bottom of menu
- [x] Move version from footer to bottom of side menu (make sure not to change too much of the html since we populate this with astringreplace)

### logically:
- [x] hide empty menus (menus without an entry)
- [x] sort entries alphabetically (in order to achieve this probably load all menu points dynamically)
- [x] link HOME to / (Default view for root is configurable so / will just go to the correct home view)
- [x] plans-link does not bring you to the plans-list
- [x] Fold open the menu section that is currently open (so that contains current view)
 
### code:
- [x] cleanup: remove sandbox files (whole `root-new-acco` folder)
- [x] migrate sidebar to a NG2+ component

## more todos:
- [x] arrows to open/close menu items are not icons
- [x] I think you switched also the arrows (or our designer did) closed items should point down
- [x] We can remove the "Categories" string (It was in teh design but I think it is just a waste of space)
- [x] You can already place the user selector in the blue area and enlarge that slightly, I will send you updated designs once I have them
- [x] when scrolling the tenannt selector stays position:fixed
- [x] I would put the scrollbar only to the menu (so everything under the horizontal line, this would also fix the other issue)
- [x] Document Compare Mask should do a line break. Lets see how that looks, if we don't like it we will find an alternative solution but we should still have the css in place)
- [x] We should hide the scrollbar:
```
/* Hide scrollbar for Chrome, Safari and Opera */
.example::-webkit-scrollbar {
  display: none;
}

/* Hide scrollbar for IE, Edge and Firefox */
.example {
  -ms-overflow-style: none;  /* IE and Edge */
  scrollbar-width: none;  /* Firefox */
}
```

## even more todos:
- [x] move all CSS to component scss


chillbirdo:
Can be checked out and tried.

The following is still to be done:

## visually:
- [ ] add all icons
- [ ] change `^` sign to appropriate icon
- [ ] highlight the currently active menu point (can be done in pure CSS)

## logically:
- [ ] hide empty menus (menus without an entry)
- [ ] sort entries alphabetically (in order to achieve this probably load all menu points dynamically

Feel free to extend/edit this list.

projects/step-core/src/lib/shared/legacy-view-registry.service.ts
```diff
@@ -39,7 +40,14 @@ export abstract class LegacyViewRegistryService {
...
   ): void;
-  abstract registerCustomMenuEntry(label: string, viewId: string, menuIconsClass: string, right: string): void;
+  abstract registerCustomMenuEntry(
+    label: string,
+    viewId?: string,
+    mainMenu?: boolean,
+    menuIconsClass?: string,
+    right?: string,
+    includedInMainMenu?: string
+  ): void;
```
neogucky, line 50: I don't like this functions parameters. Looks like mainMenu is boolean | undefined | null which is weird. I think you are using this to determine if it should show in the main menu. Instead I would rename "includedInMainMenu" which probably means "parentMenu" and either use "root" (maybe even in the ng2+ code as a static variable from the service) or undefined for root menue items.

neogucky: And comes to think of it, by switching to material I would rename "menuIconsClass" to "icon", but check if menuIconsClass is used somewhere else then the menu.

chillbirdo:
- [ ] arrows to open/close menu items are not icons
- [ ] I think you switched also the arrows (or our designer did) closed items should point down
- [ ] We can remove the "Categories" string (It was in teh design but I think it is just a waste of space)
- [ ] You can already place the user selector in the blue area and enlarge that slightly, I will send you updated designs once I have them
- [ ] when scrolling the tenannt selector stays position:fixed
- [ ] I would put the scrollbar only to the menu (so everything under the horizontal line, this would also fix the other issue)
- [ ] Document Compare Mask should do a line break. Lets see how that looks, if we don't like it we will find an alternative solution but we should still have the css in place)
- [ ] We should hide the scrollbar:
/* Hide scrollbar for Chrome, Safari and Opera */
.example::-webkit-scrollbar {
  display: none;
}

/* Hide scrollbar for IE, Edge and Firefox */
.example {
  -ms-overflow-style: none;  /* IE and Edge */
  scrollbar-width: none;  /* Firefox */
}

chillbirdo:
- [ ] change icons to material
- [ ] move all CSS to component scss

neogucky:
Can you please also add the new logo so it fits with the blue background? @chillbirdo 

logotopleft

neogucky:
@chillbirdo If you haven't done so already leave the material icons out for now (except for the ^ / v icons) since we will switch to Feathericons and probably do that after the codefreeze.

projects/step-frontend/src/lib/modules/_common/components/sidebar/sidebar.component.html
```diff
@@ -0,0 +1,63 @@
...
+      style="font-size: medium"
+      ><i class="glyphicon glyphicon-cog"></i
+    ></span>
+    <span>Categories</span>
+  </div>
+  <div id="sidebar-categories-tabs">
+    <!-- MENU -->
+    <div
+      class="sidebar-collapsable-tab"
+      *ngFor="let mme of _viewRegistryService.getCustomMainMenuEntries(); index as i"
```
neogucky, line 31: This is a huge problem, you should never invoke functions from within ng2+ html code. This function will be called on every tick and is a mssive resource leak even if it just returns an array. The solution for this is using pipes.

projects/step-frontend/src/lib/modules/_common/components/sidebar/sidebar.component.html
```diff
@@ -0,0 +1,63 @@
...
+      [attr.lastSidebarMainMenuLabel]="i === mainMenuEntriesAmount"
+    >
+      <input #mainMenuCheckBox name="{{ mme.label }}" id="collapsable-tab-{{ mme.label }}" type="checkbox" />
```
neogucky, line 34: Just for code style, you should not use curly brackets to bind values to components. Since it is a string in this case it works the same but the correct way would be: [name]="mme.label"

neogucky: Also I have a strong dislike against abrviations, so instead of "mme" I would use either "mainMenuEntry" to be specific or "entry" to be unspecific.

projects/step-frontend/src/lib/modules/_common/components/sidebar/sidebar.component.html
chillbirdo: absolutely, I just sketched this in and forgot to fix it afterwards, probably I would have noticed it next time I would have seen it. Thanks for pointing it out. The most simple solution here is not a pipe but simply calling this method once in the constructor and save the result, since the entries will not change after the application has started.

projects/step-frontend/src/lib/modules/_common/components/sidebar/sidebar.component.html
chillbirdo: I agree absolutely again, me using something like `mme` is an exception, usually I got problems the other way round - with too long variable names :) . Variable names should be self explanatory.

projects/step-frontend/src/lib/modules/_common/components/sidebar/sidebar.component.ts
```diff
@@ -0,0 +1,136 @@
...
+        this._authService.gotoDefaultPage();
+        break;
+      default:
+        this._location.go('#/root/' + viewId);
+        break;
+    }
+  }
+
+  /**
+   * FIXME: communicate open/close state as output to parent component instead of selecting global dom element
```
chillbirdo, line 115: I know that this is something that should not be done in Angular: manipulating the DOM outside of the component from withing the component.

### How do we go on with this?
I suggest that as soon as our root controller is also migrated to NG2+, we can communicate the state of the sidebar (opened or closed) to the parent component as an `@Output` and react to it from there. That would be a clean solution again. I spend some time trying to do this between the sidebar.component and the root AJS controller, but could not make it work and I did not want to waste time.

### What other solutions would be possible?
- *for the left margin of the "main" div*: build a structure that makes it obsolete to adjust the left margin manually, like a parent div with `display:flex` and `flex-direction: row`, and sidebar and main div as flex children, side-by-side. I could make this work on a sandbox page, but failed to apply this to our design successfully. Again, I did not want to waste too much time.
- *for the tenant selector*: in the current solution, the tenant selector is outside of the sidebar component and is manipulated from the outside. So why not put it inside? The reason is that the `step-tenant-selection` component is build in step EE, the `sidebar` is build in step OS. While AJS it is no problem to put an html tag in the template that is not known at compile time, the NG2+ compiler screams. Also AFAIK we do not have a migrated version of Dashlet-Binding that we used in the old version, like ```				<li ng-repeat="entry in viewRegistry.getDashlets('/menu/navbar/right')" >
					<div ng-include="entry.template"></div>
				</li>```

chillbirdo: Maybe we can open a ticket for this and apply the solution suggested above as soon as the root controller is migrated.

projects/step-frontend/src/lib/modules/_common/components/sidebar/sidebar.component.ts
```diff
@@ -0,0 +1,136 @@
...
+    } else {
+      this.sideBarOpen = true;
+      document.querySelector('#main')!.classList.remove('main-when-sidebar-closed');
+      document
+        .querySelector('step-tenant-selection-downgraded')!
+        .classList.remove('tenant-selector-when-sidebar-closed');
+      console.log('open');
+    }
+  }
+}
```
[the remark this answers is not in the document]
chillbirdo: We could create a ticket for that

chillbirdo:
From a user perspective the Sidebar looks very smooth now, also the scrolling behaves well.
@neogucky I removed the arrows from the (collapsible) mainmenus when the sidebar is collapsed. I wasn't sure if you really wanted them to be there in collapsed state, in my opinion they did not look good. I think we could make a slight distinction between a main-menu icon and a submenu icon, maybe in size or color.

projects/step-frontend/src/lib/modules/_common/components/sidebar/sidebar.component.ts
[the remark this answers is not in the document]
neogucky: > From a user perspective the Sidebar looks very smooth now, also the scrolling behaves well. @neogucky I removed the arrows from the (collapsible) mainmenus when the sidebar is collapsed. I wasn't sure if you really wanted them to be there in collapsed state, in my opinion they did not look good. I think we could make a slight distinction between a main-menu icon and a submenu icon, maybe in size or color. If you want the arrows back in the collapsed version, just remove the outer span from this line: `<span><mat-icon class="chevron">chevron_right</mat-icon></span>`

Good solution to not show them in the colapsed menu, I think for now we don't need different sizes, there are small lines between the main menu segments so we will first try and see if it is enough for our users to work with it.

neogucky (dismissed):

projects/step-frontend/src/lib/angularjs/css/step-variables.css
```diff
@@ -6,4 +6,30 @@
...
+  --top-blue-border-edge-radius: 30px;
+
+  /* SIDEBAR */
+  --sidebar-width-open: 254px;
+  --sidebar-width-closed: 72px;
+
+  --sidebar-background-color: #026299;
+  --sidebar-menu-background-color: #1b72a4;
+  --sidebar-divider-color: #3582ae;
+  --sidebar-highlight-color: #3280ad;
```
neogucky, line 21: I like that you used variables, they are awesome! But please keep in mind, they are only awesome if you use them everywhere. Currently I'm changing some colours and it turns out some elements don't use the variables.

neogucky (dismissed):

projects/step-frontend/src/lib/modules/_common/components/sidebar/sidebar.component.ts
```diff
@@ -0,0 +1,119 @@
...
+import { AJS_MODULE, AuthService, ViewRegistryService, ViewStateService } from '@exense/step-core';
+import { Location } from '@angular/common';
+
+@Component({
+  selector: '[step-sidebar]',
+  templateUrl: './sidebar.component.html',
+  styleUrls: ['./sidebar.component.scss'],
+})
+export class SidebarComponent implements AfterViewInit {
+  @ViewChildren('mainMenuCheckBox') mainMenuCheckBoxes?: QueryList<any>;
```
dvladir, line 12: Should be ElementRef instead of any

projects/step-frontend/src/lib/modules/_common/components/sidebar/sidebar.component.ts
```diff
@@ -0,0 +1,119 @@
...
+    }
+  }
+
+  private openMainMenu(mainMenuName: string): void {
+    for (let checkbox of this.mainMenuCheckBoxes!) {
+      let name = checkbox.nativeElement.getAttribute('name');
+      if (name === mainMenuName) {
+        checkbox.nativeElement.setAttribute('checked', true);
+      }
+    }
```
dvladir, line 85: Why not to use find method?
```
const chechbox = this.mainMenuCheckBoxes?.find((item) => item.nativeElement.getAttribute('name') === mainMenuName);
if (checkbox) {
    checkbox.nativeElement.setAttribute('checked', true);
}
```

projects/step-frontend/src/lib/modules/_common/components/sidebar/sidebar.component.ts
```diff
@@ -0,0 +1,119 @@
...
+      'Quota Manager',
+      'gridquotamanager',
+      'glyphicon glyphicon-road',
+      'status-root'
+    );
+    this.mainMenuEntriesAmount = _viewRegistryService.getCustomMainMenuEntries().length - 1;
+  }
+
+  ngAfterViewInit(): void {
+    let initialViewName = this._viewStateService.getViewName();
```
dvladir, line 70: If variable is immutable, const usage is preferred

projects/step-frontend/src/lib/modules/_common/components/sidebar/sidebar.component.html
```diff
@@ -0,0 +1,63 @@
...
+        [attr.lastSidebarMainMenuLabel]="i === mainMenuEntriesAmount"
+      >
+        <input
+          #mainMenuCheckBox
+          [name]="mainMenuEntry.viewId"
+          id="collapsable-tab-{{ mainMenuEntry.label }}"
+          type="checkbox"
+        />
+        <label for="collapsable-tab-{{ mainMenuEntry.label }}">
+          <span [class]="mainMenuEntry.icon + ' visible-when-collapsed'"></span>
```
dvladir, line 34: Instead of concatenation in template the better approach would be
```
<span class="visible-when-collapsed" [ngClass]="mainMenuEntry.icon"></span>
```

projects/step-frontend/src/lib/modules/_common/components/sidebar/sidebar.component.html
```diff
@@ -0,0 +1,63 @@
...
+        ><i class="glyphicon glyphicon-menu-hamburger"></i></label
+    ></span>
+    <span class="sidebar-aligned-text"><img class="sidebar-logo" alt="" [src]="logo" /></span>
+  </div>
+  <div id="sidebar-home-and-settings">
+    <span class="visible-when-collapsed sidebar-home-button" (click)="navigateTo('home')"
+      ><i class="glyphicon glyphicon-home"></i
+    ></span>
+    <span class="sidebar-home-and-settings-divider">|</span>
+    <span (click)="navigateTo('admin')" *ngIf="_authService.hasRight('admin')" class="sidebar-home-and-settings-icon">
```
dvladir, line 15: Method invocation in template's binding. Replace with
```
*ngIf="('admin' | hasRight | async)"
```

projects/step-frontend/src/lib/modules/_common/components/sidebar/sidebar.component.ts
```diff
@@ -0,0 +1,119 @@
...
+        this._location.go('#/root/' + viewId);
+        break;
+    }
+  }
+
+  /**
+   * FIXME: communicate open/close state as output to parent component instead of selecting global dom element
+   */
+  toggleOpenClose() {
+    if (this.sideBarOpen) {
```
dvladir, line 103: This code works, and I think we can continue to use it, while the app is hybrid.
But I would prefer to reference to document in `angular way` by injecting it:
```
constructor(
    ...
    @Inject(DOCUMENT) private _doc: Document
) {
  ....
}

...

toggleOpenClose(): void {
   ...
   this._doc.querySelector('#main')...
   ...
}

...


```

dvladir (changes requested):
Cool work!
Some tiny improvements required.

projects/step-frontend/src/lib/modules/_common/components/sidebar/sidebar.component.html
chillbirdo: Thanks for the feedback! All done!

projects/step-frontend/src/lib/modules/_common/components/sidebar/sidebar.component.ts
neogucky: You are right. But as Gilbert mentioned I would actually prefer to change this after the migration to not use document context at all.

chillbirdo:
> Cool work! Some tiny improvements required.

Thanks for the feedback! All fixed!

projects/step-frontend/src/lib/modules/_common/components/sidebar/sidebar.component.ts
```diff
@@ -0,0 +1,120 @@
...
+})
+export class SidebarComponent implements AfterViewInit {
+  @ViewChildren('mainMenuCheckBox') mainMenuCheckBoxes?: QueryList<ElementRef>;
+  @Input() logo: string = 'images/logotopleft.png';
+
+  mainMenuEntriesAmount: number;
+  sideBarOpen: boolean = true;
+
+  constructor(
+    @Inject(DOCUMENT) private _document: Document,
```
neogucky, line 19: Please avoid abbreviations like "doc"

dvladir (dismissed):

neogucky (approved):

PR #91, opened 2022-08-24
SED-1186: Migration of Tabs to NG2+

dvladir:


neogucky:
Can be tested on https://stepos-sed-1186-tabs.exense.ch/#/root/plans/list

neogucky:
Thanks for the great work, it is exactly as planed! That said, after playing around a bit I'm not happy with the design, especially in the execution view. Our designer will take a look. Probably we will have three differently designed tab-bars:
- 3-item-tab-bar in the side menu will have your current design
- dynamic-tab-bar might keep the design but might also get a different one
- the multi-item-tab-bar (for example the 4 items for the execution-details-tab) will get a different design (probably standard material tab-bar with underlines to mark selected tab)

Can you set the PR as draft until we have a decision for this?

chillbirdo (approved):
clean code, design needs to be improved, especially execution-tabs

PR #95, opened 2022-08-31
SED-1226 check if version available otherwise n/a

chillbirdo:
for all the functionPackages from the server that I have observed so far, `packageAttributes` was null


chillbirdo:
in case you want to recreate the bug, simply go to https://stepee-master.exense.ch/#/root/functionPackages?tenant=Common and open the developer console

chillbirdo:
image
the errors were also resulting in unrendered buttons at the right

neogucky (approved):

PR #100, opened 2022-09-05
SED-1238: Tables are not refreshed

dvladir:


neogucky (approved):
We will have to try it out and see how good this behaves

PR #108, opened 2022-09-06
SED-1280 introduce configurable step-spinner component

chillbirdo:


chillbirdo:
Suggested Vlad as reviewer, since it is related to tables

projects/step-core/src/lib/modules/table/components/table/table.component.html
```diff
@@ -1,6 +1,4 @@
-<div class="spinner-container" *ngIf="(tableDataSource?.inProgress$ | async) || inProgress">
-  <mat-spinner></mat-spinner>
-</div>
+<step-spinner [show]="!!((tableDataSource?.inProgress$ | async) || inProgress)" [showOnlyOnce]="true"></step-spinner>
```
dvladir, line 1: It means that spinner will be displayed only at first load. Correct?

1) I don't think that it's a correct approach - as task's description mention the auto-refresh.
Pagination, search, sorting - in my opinion spinner should remain for them.

2) Table - is a common reusable component. It's logic should be simple - I would keep it here is as:
if flag is true - show spinner, if flag is false - hide spinner.
For me more elegant would be to solve this problem in the data-source level.
For example we have a `reload` method in datasource. 
What if it will accept an argument `reload(showProgress: boolean = true): void`. 
If it's set as `false` datasource won't update it's _inPgoress$ subject.
Then in all autorefresh invocations, you will just have to invoke `reload(false)`

neogucky (changes requested):
See Vladimirs comment

projects/step-core/src/lib/modules/table/components/table/table.component.html
chillbirdo: 1) pagination, search, sorting would not show a spinner as well, that's correct
2) agree, this is why I didn't touch the table logic in this approach and created a spinner with its own, very simple logic

At least for now the downside of the tablesource approach suggested by @dvladir is that it does not cover `table-local-data-source`. In such cases as this:  https://github.com/exense/step-frontend/blob/0b11fb60e7a3eac67b3909c84ccc8bd00df148f8/projects/step-frontend/src/lib/modules/admin/components/screen-configuration-list/screen-configuration-list.component.ts#L23 , the TableSource is not connected to the `inProgress` flag which controls the visibility of the spinner at all. It is true that there exists a ticket that aims to avoid such cases:  https://exense.atlassian.net/browse/SED-1145

I implemented the suggested solution in https://github.com/exense/step-frontend/pull/110 .

neogucky: @chillbirdo does it still make sense to merge this one in or is everything solved by https://github.com/exense/step-frontend/pull/110 ?

chillbirdo: No this branch can be erased :cry:

neogucky: Please remember to move PRs that should not be merged to master into Draft (if you don't want to close it to still discuss something) so nobody (me) merges something that should not be merged by accident.

PR #111, opened 2022-09-06
SED-1305 Prevent sending empty testcases array

neogucky:


projects/step-frontend/src/lib/modules/execution/shared/keyword-parameters.ts
```diff
@@ -3,5 +3,5 @@ import { TableParameters } from '@exense/step-core';
 export interface KeywordParameters extends TableParameters {
   readonly type: string;
   readonly eid?: string;
-  readonly testcases: ReadonlyArray<string>;
+  readonly testcases: ReadonlyArray<string> | null;
```
dvladir, line 6: Should we explicitly send `null`, or it can be send as undefined?
In case of undefined - we can just mark field with `?`

dvladir (dismissed):

dvladir (approved):

PR #113, opened 2022-09-07
SED-1291 mapped export to new async api

neogucky:


projects/step-frontend/src/lib/angularjs/js/export.js
```diff
@@ -23,16 +23,14 @@ angular
     var factory = {};
 
     factory.pollUrl = function (exportUrl, callback) {
-      var pollCount = 0;
       (function poll() {
```
dvladir, line 26: @neogucky 
When I moved this logic in the migrated code to `poll-async-task.ts`, I've also added the maximum limit of poll counts.
But I didn't notice, that it wasn't properly used here.

Should we remove the limit in `poll-async-task.ts` too? Personally I think it's a good idea to have a limitation, but probably the default amount of maximum polls could be increased.

dvladir (dismissed):

projects/step-frontend/src/lib/angularjs/js/export.js
neogucky: I just removed it here since it was not used. I'm not convinced that a limit makes sense, because I can imagine having async operations that take hours which would be fine.

I just looked into poll-async-task.ts and it looks like this will fail every time an async operation takes longer than 2 seconds, is this correct?

dvladir: Not exactly.

It take a timeout for 500ms between each request.
The request's time also counts.

By default it will perform maximum 4 requests sequentially.

dvladir (approved):

PR #117, opened 2022-09-08
SED-1290 global entity icon missing in tables

neogucky:


projects/step-frontend/src/lib/angularjs/js/controllers/plans.js
```diff
@@ -30,7 +30,8 @@ angular
       'st-table',
       '/partials/plans/planSelectionTable.html',
       null,
-      'glyphicon glyphicon-file'
+      'glyphicon glyphicon-file',
+      'description'
```
dvladir, line 34: Should we keep this legacy registration?
For migrated tables, entities are registered in ng2+ modules

neogucky: You are right, we can remove the ones that are migrated already, I will also remove the old icon since it is not used anywhere anymore.

dvladir (approved):

PR #118, opened 2022-09-08
SED-1320 only throw one error per missing artefact

chillbirdo:


projects/step-frontend/src/lib/modules/plan/services/artefact.service.ts
```diff
@@ -29,21 +31,23 @@ export class ArtefactService {
...
+  private getType(typeName: string): ArtefactType | undefined {
     if (this.registry[typeName]) {
       return this.registry[typeName];
-    } else {
-      throw 'Unknown artefact type ' + typeName;
+    } else if (!this.missingArtefactErrorThrown[typeName]) {
+      this.missingArtefactErrorThrown[typeName] = true;
+      throw 'Unknown artefact type "' + typeName + '". Please add it to the type-registry.';
     }
+    return undefined;
```
neogucky, line 50: We should not add extensive logic to handle dev errors. If there should be an unknown artefact type it is fine to flood the console with it.

neogucky (changes requested):
Propose not to merge but I will leave it open for discussion.

projects/step-frontend/src/lib/modules/plan/services/artefact.service.ts
chillbirdo: Are you sure? I think flooding the console costs a lot of performance and in this case the app is not broken, but there is only an unusable artifact. So if one of our clients ever get into this situations, the app is still usable without any performance issues.

neogucky: Still we would inject extremely specific code to prevent errors that should never happen. If someone creates a custom plugin that triggers this error it is actually a good thing, that he will notice it for sure when looking into the console.

PR #119, opened 2022-09-08
SED-1302 filtering for interactive session id and switch table on stop

chillbirdo:


neogucky:
@chillbirdo closed as discussed in the ticket

PR #121, opened 2022-09-12
SED-1323 basic styling for AJS components

neogucky:


projects/step-app/src/styles.scss
```diff
@@ -1,5 +1,198 @@
 @use 'projects/step-frontend/styles/common-mixins';
 @import 'projects/step-frontend/styles/fonts-setup';
 
+/* Design to align with new step layout */
```
dvladir, line 4: I think it would be nicer to put these rules to as a part of the theme mixin.
So it could be applied in both versions and use the same configuration based on scss variables.

dvladir (changes requested):
Form new styles rules as one of the parts of _theme-config.scss

PR #122, opened 2022-09-12
SED-1233: Design + Material theme

dvladir:
1. Theme configuration was moved to scss mixins, to make it similar with angular material configuration
2. Theme can be configured by special components, which can be provided as plugin (see `default-theme` component & `default-theme` module for example)
3. side-bar and top-bar are customizable now
4. Color and fonts were aligned to design
5. icons and component's layout will be changed in the separate task


chillbirdo:
Looks good, powerful way to structure styles.
Scrolling works perfectly, no glitches as if https://github.com/exense/step-frontend/pull/127 would have been merged into it.

Some remarks:
- Submenu "Document Compare Masks" has no more line break it in. 
image
using `white-space: normal; line-height: 16px` it looks like that:
image
Not sure whether we should not make the font smaller. Looking at the Template, the font looks a bit too lange and thin anyways

chillbirdo (dismissed):

neogucky (dismissed):
Looks great!

@chillbirdo The Document Compare Masks will be renamed to "Compate Masks" as to avoid line breaks. Later we should probably fix it but I'm not yet sure what the best fix would be. 

The font matches the Figma design (Robot, 16px, weight:400), I think we will look at it together with the designer before release and might do some minimal adjustments.

PR #123, opened 2022-09-12
SED-1302 SED-1304 dealing with resolvedArtefact, add init declaration…

chillbirdo:
… to ngOnChanges


dvladir (approved):

neogucky (changes requested):
I will check if it makes sense to change the _class attribute on API level in the BE or to do it here. In general I liked the full class better than the new values.

PR #124, opened 2022-09-13
SED-1281 Fixed naming for bulk operations

neogucky:
Small naming changes to make operations more clear

### DIFF
--- a/projects/step-app/src/styles.scss
+++ b/projects/step-app/src/styles.scss
@@ -103,7 +103,3 @@ body {
     font-size: 18px;
   }
 }
-#main > div {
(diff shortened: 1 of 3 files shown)


projects/step-app/src/styles.scss
```diff
@@ -103,3 +103,7 @@ body {
     font-size: 18px;
   }
 }
+#main > div {
```
chillbirdo, line 106: can be removed since solved in https://github.com/exense/step-frontend/pull/125

chillbirdo (changes requested):

dvladir (approved):

PR #126, opened 2022-09-13
SED-1300 filtering for interactive session id and switch table on stop

neogucky:


chillbirdo (approved):
sorry for the wrong naming!

PR #128, opened 2022-09-13
SED-1315 set max width for tenant chooser

chillbirdo:
image


neogucky (approved):
This solves it, the paddings can be adjusted in another PR

PR #129, opened 2022-09-14
SED-1305 table keyword calls in execution details is empty

neogucky:


projects/step-frontend/src/lib/modules/execution/services/executions-panels.service.ts
```diff
@@ -34,42 +38,65 @@ export class ExecutionsPanelsService {
...
+    if (this._panels[eid]) {
+      return this._panels[eid][viewId];
+    }
+    return this._defaultPanels[viewId];
   }
 
-  getPanel(viewId: string): ExecutionStepPanel | undefined {
-    return this._panels[viewId];
+  observePanel(viewId: string, eid: string): Observable<ExecutionStepPanel | undefined> {
+    if (!this._panels$[eid] || !this._panels$[eid][viewId]) {
```
dvladir, line 53: Can be shorted as
`!this._panels$?.[eid]?.[viewId]`

projects/step-frontend/src/lib/modules/execution/components/execution-step/execution-step.component.ts
```diff
@@ -43,16 +43,16 @@ export class ExecutionStepComponent implements OnChanges, OnDestroy {
...
-      this.ID_PARAMETERS,
-    ].map((id) => _panelService.getPanel(id));
+  panelTestCases?: ExecutionStepPanel;
+  panelSteps?: ExecutionStepPanel;
+  panelParameters?: ExecutionStepPanel;
+
+  constructor(public _panelService: ExecutionsPanelsService) {}
+
+  ngOnInit() {
+    this._panelService.observePanel(this.ID_TEST_CASES, this.eId).subscribe((panel) => (this.panelTestCases = panel));
```
dvladir, line 53: Before each `.subscribe` add
`.pipe(takeUntil(this.terminator$))`

projects/step-frontend/src/lib/modules/execution/services/executions-panels.service.ts
neogucky: Thank you, I was always wondering why we can't do this but did not know about "?."

PR #130, opened 2022-09-14
SED-1240: feather icons

dvladir:
1. Feather icons was updated to the latest version
2. Custom icon wrapper & generator were added instead of `angular-feather`


dvladir:
1. Material icons were replaced with feather icons.
2. Dependency to material icons font was removed.

neogucky:
The icons look too big to me, I played around and would say a height of 20px for the icons in the menu and the table looks better, also the menu lables have to be moved slightly to the left when making the icons smaller.
Screenshot 2022-09-19 120650

dvladir:
1. Icon size war set as 20px
2. Had to add angularJS directive, that renders feather icons - because we have an execution tabs, that are still on angularJS but displayed in a new design.

neogucky (approved):
Nice PR!

PR #131, opened 2022-09-14
SED-1211  implement synthetic monitoring view

lucianbaciu:


projects/step-frontend/src/lib/modules/timeseries/performance-view/table/timeseries-table.component.scss
```diff
@@ -9,3 +9,11 @@
     margin-right: 12px;
   }
 }
+
+.ts-table {
+  ::ng-deep {
```
dvladir, line 14: As I know `::ng-ddep` is deprecated.
https://angular.io/guide/component-styles#deprecated-deep--and-ng-deep
It is better to avoid it's usage.

For your case - is it possible to add a class to the cell and specify the cell's height?
In that case row's height will align to cell's height.

dvladir (approved):
Haven't notice anything that might affect on other app's UI functionality.
Just have one note about `::ng-deep` usage

projects/step-frontend/src/lib/modules/timeseries/performance-view/table/timeseries-table.component.scss
neogucky: In case ::ng-deep needs to be used encapsulated it with :host, but I agree with @dvladir that we should try to avoid it.

projects/step-frontend/src/lib/modules/timeseries/uplot/uPlot.utils.ts
```diff
@@ -49,7 +59,7 @@ export class UPlotUtils {
       let lftDelta = cursorXVal - lftVal;
       let rgtDelta = rgtVal - cursorXVal;
 
-      hoveredIdx = lftDelta <= rgtDelta ? nonNullLft : nonNullRgt;
+      return lftDelta <= rgtDelta ? nonNullLft! : nonNullRgt!;
```
neogucky, line 62: Probably something you want to refactor later but just to highlight what we should improve for a better code style:

- shortened variables (Lft instead of Left)
- == instead of === (!==)
- null feels weird to me, are we sure that it is not just undefined?
- if statement without curly brackets "if (seriesData[i] != null) nonNullRgt = i;"
- statements that are difficult to read should get round brackets: "let rgtVal = (nonNullRgt == null) ? Infinity : xValues[nonNullRgt];"
- the bottom three let should be const

neogucky (changes requested):
Can you quickly check if the ::ng-deep statement still works with putting :host before it? With that change I will approve and merge.

See: https://blog.angular-university.io/angular-host-context/#thengdeeppseudoclassselector

projects/step-frontend/src/lib/modules/timeseries/uplot/uPlot.utils.ts
lucianbaciu: I agree. I will refactor this. This is also 'part' of the library, that's why it looks like this.

neogucky (approved):

PR #132, opened 2022-09-15
SED-1118 table misaligns

chillbirdo:
Cells squeeze until the content of the cells (or headers) does not allow further squeezing. All elements, including pagination and top menu are on screen.
image

.. the the horizontal scrollbar appears when the content can not shrink/break any further

image


chillbirdo:
This view made we want to add a space in between the columns:
image

However the only solution I found has a little impact on the design, so I created another branch on top and a draft PR:
https://github.com/exense/step-frontend/pull/133

chillbirdo:
Btw playing around with the execution view (see screenshot above), the status bar feels bulky as it does not change its size. There is an issue open for its style (it should be green, not white/empty as in the screenshot): https://exense.atlassian.net/browse/SED-1171 . Maybe we should include in this ticket that it should be shrinkable as well.

projects/step-core/src/lib/modules/table/components/table/table.component.scss
```diff
@@ -7,12 +11,12 @@ table {
   width: 100%;
 }
 
-tr.mat-row:hover {
-  background: var(--darken-background-10);
+::ng-deep .mat-form-field-infix {
```
neogucky, line 14: Did you check if it is possible to implement this without ::ng-deep? If not, at least add :host before it so it does not affect the whole app.

neogucky (changes requested):
I like the simplicity of the PR, but please make sure that the tables look (nearly) identical to the current design.
- no gaps in the grey lines
- paddings between cells when the space allows it

See:
Screenshot 2022-09-15 at 13 16 48

Screenshot 2022-09-15 at 13 16 59

projects/step-core/src/lib/modules/table/components/table/table.component.scss
chillbirdo: I tried everything I could think of before using `::ng-deep`. Putting it into `:host` makes perfectly sense, just did it :+1:

chillbirdo: The gaps in the lines are reason why I created an extra PR: https://github.com/exense/step-frontend/pull/133 so that you can decide what you like more.

Actually, **this** PR (132) does not produce any gaps.

chillbirdo: the only drawback is that I could not achieve to add a padding only to the content of the cells. so when the table is squeezed, there is no space between cells (see screenshot above). so either no space between cells, or the look of #133 .

neogucky: Actually this PR (#132) has the issue with the gap, I did not look at the other one. The gap is a no-go, but we also need to have some padding. 

Just a quick look for solutions:
remove 
`border-spacing: 10px 0!important;` for tr
add 
```
td.mat-cell {
    padding: 0 8px;
}
```

Probably you would have to be a bit more specific which cells should get padding (maybe not the checkbox) and maybe have to set the paddign according to viewport.

Second solution is to simply remove `border-collapse: separate!important;` for the table. I don't see why we have this set but maybe check if there is any problem with removing it.

neogucky (approved):

projects/step-core/src/lib/modules/table/components/table/table.component.scss
```diff
@@ -1,46 +1,55 @@
...
-.spinner-container {
-  position: absolute;
-  z-index: 1;
-  width: 100px;
-  height: 100px;
-  left: calc(50% - 50px);
-  top: calc(50% - 50px);
-}
+  tr.mat-row:hover {
+    background: var(--darken-background-10);
```
dvladir, line 23: As css variables are no longer used in ng2+ part, this rule is redundant

projects/step-core/src/lib/components/array-filter/array-filter.component.scss
```diff
@@ -0,0 +1,3 @@
+@use '../../../../styles/core-mixins' as core;
```
dvladir, line 1: I think that absolute import is better, than relative here, because if you will decide to move component somewhere else in future, you should keep in mind to change this relative import too.
In case of absolute import, while component in the same project the import path will always be the same
`@use 'projects/step-core/styles/core-mixins' as core;`

projects/step-core/src/lib/modules/table/components/table/table.component.scss
```diff
@@ -1,46 +1,55 @@
...
-  ::ng-deep step-btn-group {
-    opacity: 0.5;
+  mat-paginator {
+    border-radius: 0 0 16px 16px;
   }
-  &:hover ::ng-deep step-btn-group {
-    opacity: 1;
+
+  tr {
+    ::ng-deep step-btn-group {
```
dvladir, line 48: ng-deep usage

projects/step-core/src/lib/modules/table/components/table/table.component.scss
neogucky: @dvladir Is there already a hover effect when we remove it?

dvladir (changes requested):
Check comments above

projects/step-core/src/lib/modules/table/components/table/table.component.scss
dvladir: @neogucky As I see in UI there is no hover effect at all now. Is it a bug?

projects/step-core/src/lib/modules/table/components/table/table.component.scss
neogucky: This is encapsuled in :host so it is fine for now. When we have more time we should take a better look what to do of ng-deeps in the project.

projects/step-core/src/lib/modules/table/components/table/table.component.scss
neogucky: Seems like a bug, probably because the variable doesn't exist anymore

chillbirdo: removed it, still no hover effect

dvladir (approved):

PR #135, opened 2022-09-20
SED-1198 Filtering not working as expected

dvladir:


neogucky (approved):
I like how simple the solution is, it shows how good our architecture is set up!

PR #136, opened 2022-09-20
SED-1309 The search by Keyword type on the Keyword table doesn't work properly

dvladir:
To pass the right keyword type in request properly, I've change the search column from the free text to the dropdown filter, where all registered keyword types are displayed as dropdown's items.


neogucky (approved):
Perfect, thanks for solving the template function!

PR #138, opened 2022-09-20
SED-1318 Icon execution type missing on execution table

dvladir:


neogucky (approved):
Nice! Closing also SED-1332

PR #141, opened 2022-09-20
SED-1279 adds target blank

chillbirdo:


neogucky:
Can be closed since issue was solved in SED-1301

PR #143, opened 2022-09-21
SED-1342-Chart "Response Times" displays wrong value for AVG

lucianbaciu:
SED-1343-Measurement filters not reacting
SED-1348-Summary table of performance view doesn't react to timerange selection

### DIFF
--- a/projects/step-frontend/src/lib/modules/timeseries/performance-view/chart-generators/chart-generators.ts
+++ b/projects/step-frontend/src/lib/modules/timeseries/performance-view/chart-generators/chart-generators.ts
@@ -1,0 +1,99 @@
+import { FindBucketsRequest } from '../../find-buckets-request';
+import { TimeSeriesUtils } from '../../time-series-utils';
+import { UPlotUtils } from '../../uplot/uPlot.utils';
+import { TimeSeriesChartResponse } from '../../time-series-chart-response';
+import { TSChartSettings } from '../../chart/model/ts-chart-settings';
+import { TsChartType } from '../ts-chart-type';
+
+declare const uPlot: any;
+
+export class ChartGenerators {
+  private static readonly CHART_LEGEND_SIZE = 65;
+
+  private static readonly barsFunction = uPlot.paths.bars; // this is a function from uplot which allows to draw bars instead of straight lines
+  private static readonly stepped = uPlot.paths.stepped; // this is a function from uplot wich allows to draw 'stepped' or 'stairs like' lines
+
+  static generateChart(
+    type: TsChartType,
+    request: FindBucketsRequest,
+    response: TimeSeriesChartResponse
+  ): TSChartSettings {
+    switch (type) {
+      case TsChartType.OVERVIEW:
+        return this.createSummaryChartSettings(request, response);
+      case TsChartType.BY_STATUS:
+      case TsChartType.RESPONSE_TIME:
+      case TsChartType.THROUGHPUT:
+      case TsChartType.THREAD_GROUP:
+        throw 'Not implemented';
+    }
+  }
+
+  static createSummaryChartSettings(request: FindBucketsRequest, response: TimeSeriesChartResponse): TSChartSettings {
+    let xLabels = TimeSeriesUtils.createTimeLabels(response.start, response.end, response.interval);
+    let avgValues: (number | null)[] = [];
+    let countValues: (number | null)[] = [];
+    if (response.matrixKeys.length === 0) {
+      // empty data
+      // if (this.summaryChart) {
+      //   this.summaryChart.clear();
+      //   return;
+      // }
+    } else {
+      response.matrix[0].forEach((bucket) => {
+        avgValues.push(bucket ? Math.trunc(bucket.sum / bucket.count) : null);
+        countValues.push(bucket?.throughputPerHour);
+      });
+    }
+
+    return {
+      title: 'Performance Overview',
+      showLegend: true,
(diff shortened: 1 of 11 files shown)


projects/step-frontend/src/lib/modules/timeseries/performance-view/chart-generators/chart-generators.ts
```diff
@@ -0,0 +1,99 @@
...
+    let countValues: (number | null)[] = [];
+    if (response.matrixKeys.length === 0) {
+      // empty data
+      // if (this.summaryChart) {
+      //   this.summaryChart.clear();
+      //   return;
+      // }
+    } else {
+      response.matrix[0].forEach((bucket) => {
+        avgValues.push(bucket ? Math.trunc(bucket.sum / bucket.count) : null);
```
jeromecomte, line 44: Is there a reason for doing this in the FE?

projects/step-frontend/src/lib/modules/timeseries/performance-view/chart-generators/chart-generators.ts
```diff
@@ -0,0 +1,99 @@
...
+          //   return gradient;
+          // },
+          width: 2,
+          stroke: 'rgba(255,109,18,0.59)',
+        },
+        {
+          id: 'count',
+          scale: 'total',
+          label: 'Hits/h',
+          data: countValues,
```
jeromecomte, line 75: I assume that the variable refer to tph and not count.

projects/step-frontend/src/lib/modules/timeseries/performance-view/chart-generators/chart-generators.ts
```diff
@@ -0,0 +1,99 @@
...
+  private static readonly barsFunction = uPlot.paths.bars; // this is a function from uplot which allows to draw bars instead of straight lines
+  private static readonly stepped = uPlot.paths.stepped; // this is a function from uplot wich allows to draw 'stepped' or 'stairs like' lines
+
+  static generateChart(
+    type: TsChartType,
+    request: FindBucketsRequest,
+    response: TimeSeriesChartResponse
+  ): TSChartSettings {
+    switch (type) {
+      case TsChartType.OVERVIEW:
```
jeromecomte, line 22: At some point we could have registry mechanism where custom charts register themselves. This would allow plugins to register custom charts and custom dashboards to reference them

projects/step-frontend/src/lib/modules/timeseries/performance-view/time-selection/execution-page-time-selection.component.ts
```diff
@@ -100,6 +104,7 @@ export class ExecutionPageTimeSelectionComponent implements OnInit {
   }
 
   prepareSelectForRanger(selection: ExecutionTimeSelection): TSTimeRange | undefined {
+    console.log(selection);
```
jeromecomte, line 107: Console.log should be removed from the code before releasing

jeromecomte (approved):
Can be merged to master in the current state. Please create tickets for the mentioned comments

projects/step-frontend/src/lib/modules/timeseries/performance-view/chart-generators/chart-generators.ts
```diff
@@ -0,0 +1,99 @@
...
+          label: 'Response Time',
+          data: avgValues,
+          value: (x, v) => Math.trunc(v) + ' ms',
+          // stroke: 'rgba(0,117,187,0.41)',
+          // fill: (self, idx) => {
+          //   let gradient = self.ctx.createLinearGradient(0, 0, 0, 400);
+          //   gradient.addColorStop(0, 'rgba(49,116,197,0.73)');
+          //   gradient.addColorStop(1, '#ff000006');
+          //   return gradient;
+          // },
```
neogucky, line 67: In general we shouldn't commit commented out code. Either it was already there and we might need it in the future, then we use the git history, or it is new code we just don't wish to include just yet, then we should stash it / leave it local.

projects/step-frontend/src/lib/modules/timeseries/performance-view/time-selection/execution-page-time-selection.component.ts
neogucky: In general we should remove console.log always before merging into master, for debug we usually can use break points. If this is not possible at all set a descriptive lable like:
` console.log('prepared ranger selection is:', selection);`

neogucky (changes requested):
Looks good to me, but I see that it is not building: Execution: 632b23bbfd5f6139bc9dadd0 

Can you check if it is building locally @lucianbaciu ? npm run build:os

projects/step-frontend/src/lib/modules/timeseries/performance-view/chart-generators/chart-generators.ts
lucianbaciu: Not really. just the fact that avg is not generated in the buckets nor service

projects/step-frontend/src/lib/modules/timeseries/performance-view/chart-generators/chart-generators.ts
lucianbaciu: yes. the other is avg, this is the volume/count/tph

neogucky:
This branch was merged as https://github.com/exense/step-frontend/pull/150 due to problems with branch name and build server.

PR #149, opened 2022-09-21
SED-1318 Showing no entity icon if entity has no icon

neogucky:


projects/step-core/src/lib/modules/entity/components/entity-icon/entity-icon.component.scss
```diff
@@ -3,3 +3,9 @@ mat-icon {
   vertical-align: middle;
   padding: 3px;
 }
+
+span {
+  width: 20px;
+  height: 20px;
+  display: inline-block;
+}
```
dvladir, line 11: I think it might be solved in more elegant way, like
```
:host {
  min-width: 20px;
  min-height: 20px;
  display: inline-block;
}
```
In that case all additional logic with *ngIf in template can be removed.

`mat-icon` rule can be removed also, as there are no more material icons

dvladir (changes requested):
See comments above

PR #150, opened 2022-09-21
Sed 1029 reimplement viz rtm bug fixing

lucianbaciu:


neogucky (approved):
Aproved as described in https://github.com/exense/step-frontend/pull/143

The original branch name had a naming problem and could not be deployed on the build server, this one builds correctly.

PR #153, opened 2022-09-22
SED-1341 adds timestamp-suffix to filename of legacy scripts

chillbirdo:
The JS files that are loaded by the browser are those:
image

The NG2+ part of the app is inside the `main.{hash}.js` which itself loads all the js files that only have numbers and a hash like `4650.e8ef9409373e0de2.js`. So the new part alread uses hashes in the filenames, which forces the browser to load it freshly.

The problem is the legacy part of the app, which renders into the `scripts.js` file. There is no hash attached to that file, so browsers may take the one that is cached. I found no way to tell the builder to attach one (also `ng build --output-hashing=all` didn't help).

So I attached a script to the `npm build` command. This script has no relevance when the app is developed locally, where the FE project is inside the workspace. It only executes on the stand-alone build on the server. It adds a timestamp to the `scripts.js` file and references to that from `index.html`.

image


dvladir (dismissed):

chillbirdo:
This script is working for OS when executed after `npm run build`. It would also work for EE when executed after the build of both, OS and EE.

Since this is an '.sh' script, DevOps ( @jeromebrongniart ) could probably use it:
```
#!/bin/bash

## add timestamp to dist/step-app/app/scripts.js and refer to it in dist/step-app/index.html in order to prevent caching of legacy code
TIMESTAMP=$(date +%s)
SCRIPTSJS_FILENAME_BEFORE=scripts.js
SCRIPTSJS_FILENAME_AFTER=scripts-$TIMESTAMP.js

SCRIPTSJS_FILE_FOLDER=dist/step-app/app
INDEX_HTML_FILE_PATH=dist/step-app/index.html

mv $SCRIPTSJS_FILE_FOLDER/$SCRIPTSJS_FILENAME_BEFORE $SCRIPTSJS_FILE_FOLDER/$SCRIPTSJS_FILENAME_AFTER
sed -i "s/$SCRIPTSJS_FILENAME_BEFORE/$SCRIPTSJS_FILENAME_AFTER/g" $INDEX_HTML_FILE_PATH
```

PR #156, opened 2022-09-26
SED-1213-TS - Implements migration from measurements to buckets colle…

lucianbaciu:
…ction

### DIFF
--- a/projects/step-frontend/src/lib/modules/timeseries/execution-page/execution-page.component.ts
+++ b/projects/step-frontend/src/lib/modules/timeseries/execution-page/execution-page.component.ts
@@ -1,5 +1,12 @@
 import { Component, Input, OnDestroy, OnInit, ViewChild } from '@angular/core';
-import { AJS_MODULE, DashboardService, ExecutionsService } from '@exense/step-core';
+import {
+  AJS_MODULE,
+  AsyncTasksService,
+  AsyncTaskStatusResource,
+  DashboardService,
+  ExecutionsService,
+  pollAsyncTask,
+} from '@exense/step-core';
 import { TimeSeriesConfig } from '../time-series.config';
 import { downgradeComponent, getAngularJSGlobal } from '@angular/upgrade/static';
 import { PerformanceViewSettings } from '../performance-view/performance-view-settings';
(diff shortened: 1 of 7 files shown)


projects/step-frontend/src/lib/modules/timeseries/util/timeseries-colors-pool.ts
```diff
@@ -51,11 +52,16 @@ export class TimeseriesColorsPool {
    */
   getStatusColor(status: string): string {
     if (!status) {
-      let randomColor = this.randomRGBA();
-      statusColors[status.toLowerCase()] = randomColor;
-      return randomColor;
+      let grey = '#cccccc';
+      statusColors[this.NO_STATUS] = grey;
+      return grey;
```
neogucky, line 57: You are setting something from within the setter. I think a cleaner solution would be to set the colour statusColors[this.NO_STATUS] = grey; in an init method and just return statusColors[this.NO_STATUS].

Also note that grey should be a const here since it is not changing.

projects/step-frontend/src/lib/modules/timeseries/execution-page/execution-page.component.ts
```diff
@@ -70,6 +92,30 @@ export class ExecutionPageComponent implements OnInit, OnDestroy {
     });
   }
 
+  rebuildTimeSeries() {
+    this.migrationInProgress = true;
+    this.timeSeriesService
+      .rebuildTimeSeries(this.executionId)
+      .pipe(
+        pollAsyncTask(this._asyncTaskService),
+        delay(10_000) // in order for the migration to be completed as much as possible
```
neogucky, line 101: I'm confused about the delay, the pollAsyncTask should have its internal 500ms delay and only return if the task is ready.

projects/step-frontend/src/lib/modules/timeseries/util/timeseries-colors-pool.ts
lucianbaciu: agree

projects/step-frontend/src/lib/modules/timeseries/execution-page/execution-page.component.ts
dvladir: Agree. delay is redunant here

neogucky (changes requested):
The delay needs to be refactored, would be great if we could also strip the comments / console.logs.

neogucky (approved):

PR #159, opened 2022-09-26
SED-1366 the charts of the perforamnce view remain empty after filter change

lucianbaciu:


projects/step-frontend/src/lib/modules/timeseries/execution-page/execution-page.component.ts
```diff
@@ -70,6 +92,30 @@ export class ExecutionPageComponent implements OnInit, OnDestroy {
...
   }
 
+  rebuildTimeSeries() {
+    this.migrationInProgress = true;
+    this.timeSeriesService
+      .rebuildTimeSeries(this.executionId)
+      .pipe(
+        pollAsyncTask(this._asyncTaskService),
+        delay(10_000) // in order for the migration to be completed as much as possible
+      )
```
neogucky, line 102: Same as in the other PR, make sure that the delay really makes sense

projects/step-frontend/src/lib/modules/timeseries/performance-view/performance-view.component.ts
```diff
@@ -240,16 +215,25 @@ export class PerformanceViewComponent implements OnInit, OnDestroy {
...
   reconstructAllCharts() {
-    this.findRequest.numberOfBuckets = this.calculateIdealNumberOfBuckets(
-      this.settings.startTime,
-      this.settings.endTime
-    );
+    this.findRequest = this.prepareFindRequest(this.settings); // we don't want to lose active filters
+    // this.findRequest.numberOfBuckets = this.calculateIdealNumberOfBuckets(
+    //   this.settings.startTime,
+    //   this.settings.endTime
+    // );
```
neogucky, line 224: I wouldn't commit commented out code to master

projects/step-frontend/src/lib/modules/timeseries/ranger/ts-ranger.component.ts
```diff
@@ -88,7 +95,9 @@ export class TSRangerComponent implements OnInit, AfterViewInit, OnChanges {
 
   selectRange(fromTimestamp?: number, toTimestamp?: number) {
     let select = this.transformRangeToSelect({ from: fromTimestamp, to: toTimestamp });
-
+    // select!.width = select!.width! + 100;
+    // console.log(this.getSize(), select);
+    console.log(this.uplot.bbox, select);
```
neogucky, line 100: I don't want to nag you all the time about debug / commented out code, but if we remove it in the PRs we have a much easier job with keeping the master clean.

projects/step-frontend/src/lib/modules/timeseries/performance-view/measurements/measurements-picker.component.ts
```diff
@@ -17,31 +18,37 @@ export class MeasurementsPickerComponent implements OnInit {
...
     if (!this.contextId) {
       throw new Error('Context Id input must be present');
     }
     this.keywordsService = this.executionPageService.getContext(this.contextId).getKeywordsContext();
-    this.keywordsService.onKeywordsUpdated().subscribe((keywords) => {
-      this.keywords = keywords;
-      this.activeKeywords = 0;
-      Object.keys(keywords).forEach((key) => {
-        if (keywords[key].isVisible && keywords[key].isSelected) {
+    this.subscriptions.add(
```
dvladir, line 30: Just FYI, another approach to control the subscriptions lifecycle - using the `takeUntil` operator:
```
// Define a subject in the component
private terminator$ = new Subject<unknown>();

...
// Before subscription add takeUntil
this.keywordsService
   .onKeywordsUpdated()
   .pipe(
     takeUntil(this.terminator$)
   )
  .subscribe((keywords) => {
     ...
  });

...
// In destroy hook invoke and complete the subject to terminate the susbscriptions

ngOnDestroy(): void {
    this.terminator$.next();
    this.terminator$.complete();
}
```

From my point of view such approach looks cleaner and easier to use.

projects/step-frontend/src/lib/modules/timeseries/ranger/ts-ranger.component.ts
lucianbaciu: my fault. no worries. I will try to get used to check for console.logs before pushing

projects/step-frontend/src/lib/modules/timeseries/performance-view/measurements/measurements-picker.component.ts
lucianbaciu: Thank you. I also like it better. I will use this from now

neogucky (changes requested):
Delay should be removed, refactoring as of Vladimirs comment + removing comments/console.logs would be great.

lucianbaciu:
> Delay should be removed, refactoring as of Vladimirs comment + removing comments/console.logs would be great.

why remove the delay?

neogucky:
> > Delay should be removed, refactoring as of Vladimirs comment + removing comments/console.logs would be great.
> 
> why remove the delay?

Because I don't see what purpose it has since the async function already is doing a delay (and actually repeating polls until the async tasks finishes)

neogucky:
> 

Alright, but then we should change it in the BE so it becomes part of the async task OR create a second async task that is a super-set of the other tasks + the finishing one. I don't like to have a hardcoded delay. For now it is fine though and I will approve.

neogucky (dismissed):

PR #160, opened 2022-09-27
Sed 1365 unify headers

chillbirdo:
I chose a generic solution, which was in my optinion the most elegant, so that multiple table-headers are allowed and any combination of them can be combined optionally.

image
image
image


dvladir (changes requested):
@chillbirdo Really nice solution.

Just few moments to fix:
- input parameter in additional header
- setupAdditionalHeaderFunction - not mandatory, I just found it cleaner and easier to read

dvladir (approved):

neogucky (approved):

PR #163, opened 2022-09-29
SED-1400 solved incorrect permissions

neogucky:


david-stephan (changes requested):
See "suggested" changes directly in the review.
Notes: 
- we have specific rights for masks
- we have *-delete rights for most entities except users

PR #164, opened 2022-09-29
SED-1397 Executions List and Execution Details style

dvladir:


neogucky (changes requested):
Elegant and simple solutions. Requesting changes regarding some misalignment. Also Adding some more points reported by Jerome when reviewing your solution that initially where not included in the ticket but fit there. (see ticket)

neogucky (approved):
Great styling!

PR #165, opened 2022-09-29
SED-1347-Added missing units

lucianbaciu:


neogucky (approved):
Looks good! Thanks for creating the PR individually

PR #166, opened 2022-09-30
SED-1340 Solved issue preventing re-run of executions

neogucky:


chillbirdo:
this solves the issue, but `table` will always be undefined since https://github.com/exense/step-frontend/blob/fa20f6e9e93ffc02e2217b7b5ef26266a8d2a03a/projects/step-frontend/src/lib/angularjs/partials/execution/executionStep.html#L38

PR #167, opened 2022-10-03
SED-1407 break word and set width dependent to vw

chillbirdo:


neogucky (changes requested):
This doesn't really work, the width is fixed so it looks bad on smaller and bigger screens and also the line is breaking so it is in a different line than the controls name (i.e. echo in your test)

I looked a bit into it and putting the parent div to "word-break: break-word;" and removing the added lines here solves the issue. Can you test this and adjust accordingly?

neogucky (approved):
Looks good, thank you!

PR #170, opened 2022-10-03
SED-1408 adapted row height and font size

chillbirdo:
image

### DIFF
--- a/projects/step-core/src/lib/modules/table/components/table/table.component.scss
+++ b/projects/step-core/src/lib/modules/table/components/table/table.component.scss
@@ -46,6 +46,10 @@
     border-radius: 0 0 16px 16px;
   }
 
+  tr[mat-header-row] {
+    height: 63px;
+  }
+
   tr {
     ::ng-deep step-btn-group {
       opacity: 0.5;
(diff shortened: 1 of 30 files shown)


neogucky (changes requested):
Here we are changing the design for all tables to solve an edgecase. Let's just hide the lable for now until we can find a solution with our designer.

chillbirdo:
considered @dvladir comment in Jira

projects/step-core/src/lib/modules/table/components/table/table.component.scss
```diff
@@ -46,6 +46,10 @@
     border-radius: 0 0 16px 16px;
   }
 
+  tr[mat-header-row] {
+    height: 63px;
+  }
```
neogucky, line 51: This is still the incorrect row height

neogucky (changes requested):
We still have to adjust the row height. Also I would prefer to use styles in legacy-styles to hide the error because:
- I don't like commented out code checked in
- We can't forget about these styles when we start working on 3.21 (since we will delete the legacy-styles first thing)

PR #179, opened 2022-10-04
SED-1425-Frontend is using the resolution property from the backend

lucianbaciu:


projects/step-frontend/src/lib/modules/timeseries/performance-view/performance-view.component.ts
```diff
@@ -511,7 +510,10 @@ export class PerformanceViewComponent implements OnInit, OnDestroy {
    */
   calculateIdealNumberOfBuckets(startTime: number, endTime: number): number {
     return Math.trunc(
-      Math.min(TimeSeriesConfig.MAX_BUCKETS_IN_CHART, (endTime - startTime) / TimeSeriesConfig.RESOLUTION / 2)
+      Math.min(
+        TimeSeriesConfig.MAX_BUCKETS_IN_CHART,
+        (endTime - startTime) / this.timeSeriesService.getResolution() / 2
```
neogucky, line 515: I would suggest adding backets to clearify the devision sequence.

projects/step-frontend/src/lib/modules/timeseries/performance-view/time-selection/execution-page-time-selection.component.ts
```diff
@@ -78,7 +78,10 @@ export class ExecutionPageTimeSelectionComponent implements OnInit, OnDestroy {
       start: startTime,
       end: endTime, // to current time if it's not ended
       numberOfBuckets: Math.trunc(
-        Math.min(TimeSeriesConfig.MAX_BUCKETS_IN_CHART, (endTime - startTime) / TimeSeriesConfig.RESOLUTION / 2)
+        Math.min(
+          TimeSeriesConfig.MAX_BUCKETS_IN_CHART,
+          (endTime - startTime) / this.timeSeriesService.getResolution() / 2
```
neogucky, line 83: I would suggest adding backets to clearify the devision sequence.

projects/step-frontend/src/lib/modules/timeseries/time-series.service.ts
```diff
@@ -3,14 +3,21 @@ import { Injectable } from '@angular/core';
...
 @Injectable({
   providedIn: 'root',
 })
 export class TimeSeriesService {
-  constructor(private http: HttpClient) {}
+  private readonly resolutionPeriod: number; // this is received from the backend
+
+  constructor(private http: HttpClient, private authService: AuthService) {
+    this.resolutionPeriod =
+      parseInt(authService.getConf()?.miscParams[TimeSeriesConfig.SERVICE_RESOLUTION_CONFIG_PROPERTY]!) ||
```
neogucky, line 18: parseInt looks incorrect. I'm not 100% sure what it does, but in TS you would usually use the Number constructor:
Number(authService.getConf()?.miscParams[TimeSeriesConfig.SERVICE_RESOLUTION_CONFIG_PROPERTY]!)

Also it looks weird that you use ? to say getConf() might be undefined but then guarantee that there is a value with the closing !. In general using the exclamation mark without checks should be avoided except if you are 100% sure it will always work, then you can replace the ? also with one.

dvladir: @neogucky 
`parseInt` takes string and tries to parse it and return number.
But if the string was incorrect, it returns NaN.
I think it's okay to use `parseInt` here, but it's result should be additional checked with `isNaN` function

neogucky: > @neogucky `parseInt` takes string and tries to parse it and return number. But if the string was incorrect, it returns NaN. I think it's okay to use `parseInt` here, but it's result should be additional checked with `isNaN` function

I still prefer Number since parseInt is extremely error tolerant, something we shouldn't need in our own eco-system: https://thisthat.dev/number-constructor-vs-parse-int/#:~:text=Differences,parses%20the%20value%20of%20input.&text=As%20you%20see%2C%20parseInt%20will,to%20convert%20the%20entire%20string.

projects/step-frontend/src/lib/modules/timeseries/time-series.service.ts
```diff
@@ -3,14 +3,21 @@ import { Injectable } from '@angular/core';
...
+import { AsyncTaskStatusResource, AuthService, Execution } from '@exense/step-core';
+import { TimeSeriesConfig } from './time-series.config';
 
 // TODO provide it it TimeSeriesModule
 @Injectable({
   providedIn: 'root',
 })
 export class TimeSeriesService {
-  constructor(private http: HttpClient) {}
+  private readonly resolutionPeriod: number; // this is received from the backend
```
dvladir, line 14: As this field is readonly and assigned in the contstructor, `private` modifier and `getResolution` method looks redundant

PR #183, opened 2022-10-05
SED-1282 ts   add final styles (align all different elements, etc)

lucianbaciu:


projects/step-frontend/src/lib/modules/timeseries/chart/time-series-chart.component.scss
```diff
@@ -9,6 +9,23 @@
...
+    .u-series {
+      margin-right: 8px;
+      white-space: nowrap;
+
+      .u-marker:empty {
+        display: none;
+      }
+    }
+  }
+}
```
neogucky, line 27: if you really can't avoid ::ng-deep use :host to prevent bleed. Better even would be to find a solution to add the styles in the child component

projects/step-frontend/src/lib/modules/timeseries/chart/time-series-chart.component.ts
```diff
@@ -57,6 +57,13 @@ export class TimeSeriesChartComponent implements OnInit, AfterViewInit, OnChange
     }
   }
 
+  setTitle(title: string): void {
+    let titles = this.chartElement.nativeElement.getElementsByClassName('u-title');
+    if (titles.length) {
+      titles[0].innerHTML = title;
+    }
+  }
```
neogucky, line 65: In general we shouldn't need this kind of programatical way of setting things but use inputs in the template

projects/step-frontend/src/lib/modules/timeseries/chart/time-series-chart.component.scss
lucianbaciu: the child component is the component genreated by the chart library. I can go with :host instead.

projects/step-frontend/src/lib/modules/timeseries/chart/time-series-chart.component.ts
lucianbaciu: I checked the entire library for ability to change the title dinamically, but it does not have it. This is the only option in this case since I'm manipulating directly elements generated by uPlot.

projects/step-frontend/src/lib/modules/timeseries/chart/time-series-chart.component.scss
lucianbaciu: :host is not working here..

PR #185, opened 2022-10-05
SED-1438-Added all status colors

lucianbaciu:


neogucky:
I didn't go deep in your code, do you need the colours in the controller or could you use styles as well? If yes you could reuse our mixin from variables.scss:

@mixin status($status, $color) {
  .step-#{$status} {
    background: $color;
    color: #fff;
  }
}

@each $status, $color in $statuses {
  @include status($status, $color);
}

lucianbaciu:
> I didn't go deep in your code, do you need the colours in the controller or could you use styles as well? If yes you could reuse our mixin from variables.scss:
> 
> ```
> @mixin status($status, $color) {
>   .step-#{$status} {
>     background: $color;
>     color: #fff;
>   }
> }
> 
> @each $status, $color in $statuses {
>   @include status($status, $color);
> }
> ```

Hi Tim, unfortunately I can't use css here. These colors are passed in several place in typescript (also to the chart series).

neogucky (approved):

PR #186, opened 2022-10-05
SED-1433 Visualization with npm doesn't work anymore

dvladir:


neogucky (approved):
Great solution, thanks!

PR #192, opened 2022-10-11
SED-1418 Running executions query just when the previous requests…

lucianbaciu:
… are done

### DIFF
--- a/projects/step-frontend/src/lib/modules/timeseries/execution-page/execution-page.component.ts
+++ b/projects/step-frontend/src/lib/modules/timeseries/execution-page/execution-page.component.ts
@@ -14,7 +14,7 @@ import { TimeSeriesService } from '../time-series.service';
 import { TimeSeriesContextsFactory } from '../time-series-contexts-factory.service';
 import { RangeSelectionType } from '../time-selection/model/range-selection-type';
 import { PerformanceViewComponent } from '../performance-view/performance-view.component';
-import { delay, forkJoin, map, Subject, take, timer } from 'rxjs';
+import { first, forkJoin, Subject, timer } from 'rxjs';
 
 @Component({
   selector: 'step-execution-performance',
@@ -139,7 +139,7 @@ export class ExecutionPageComponent implements OnInit, OnDestroy {
   }
 
   triggerNextUpdate(delay: number) {
-    forkJoin([timer(delay), this.dashboardUpdateCompleteSubject.pipe(take(1))]).subscribe(() => {
+    forkJoin([timer(delay), this.dashboardUpdateCompleteSubject.pipe(first())]).subscribe(() => {
       let now = new Date().getTime();
       if (!this.intervalShouldBeCanceled) {
         this.performanceViewSettings!.endTime =
--- a/projects/step-frontend/src/lib/modules/timeseries/execution-page/execution-page.component.ts
+++ b/projects/step-frontend/src/lib/modules/timeseries/execution-page/execution-page.component.ts
@@ -1,13 +1,5 @@
 import { Component, Input, OnDestroy, OnInit, ViewChild } from '@angular/core';
-import {
-  AJS_MODULE,
-  AsyncTasksService,
-  AsyncTaskStatusResource,
-  DashboardService,
-  ExecutionsService,
-  pollAsyncTask,
-} from '@exense/step-core';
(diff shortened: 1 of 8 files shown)


projects/step-frontend/src/lib/modules/timeseries/performance-view/performance-view.component.ts
```diff
@@ -210,30 +220,38 @@ export class PerformanceViewComponent implements OnInit, OnDestroy {
...
-  // we assume the settings input has been changed
-  reconstructAllCharts() {
-    this.findRequest = this.prepareFindRequest(this.settings); // we don't want to lose active filters
-    this.updateAllCharts();
-  }
-
-  mergeRequestWithActiveFilters() {
-    this.deleteObjectProperties(this.findRequest.params, this.filtersComponent.getAllFilterAttributes()); // we clean the request first
-    Object.assign(this.findRequest.params, this.executionContext.getActiveFilters());
+    forkJoin(this.initializationTasks.map((obs) => obs.pipe(take(1)))).subscribe((allCompleted) =>
```
neogucky, line 226: take(1) = first() (first() has a nicer readability)

projects/step-frontend/src/lib/modules/timeseries/performance-view/performance-view.component.ts
```diff
@@ -210,30 +220,38 @@ export class PerformanceViewComponent implements OnInit, OnDestroy {
...
+      this.onInitializationComplete.emit()
+    );
   }
 
   updateAllCharts() {
+    this.updateTasks = [];
```
neogucky, line 232: Do we really need the differentiation of updateTasks and initializeTasks? I see a lot of duplications between these.

projects/step-frontend/src/lib/modules/timeseries/execution-page/execution-page.component.ts
```diff
@@ -121,47 +130,38 @@ export class ExecutionPageComponent implements OnInit, OnDestroy {
...
       this.refreshEnabled = false;
-      clearInterval(this.intervalExecution);
     }
     this.selectedRefreshInterval = newInterval;
   }
 
-  startRefreshInterval(interval: number) {
-    this.intervalExecution = setInterval(() => {
+  triggerNextUpdate(delay: number) {
+    forkJoin([timer(delay), this.dashboardUpdateCompleteSubject.pipe(take(1))]).subscribe(() => {
```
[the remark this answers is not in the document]
neogucky: As Discussed the mayor issue is not a problem, we still should try to improve the architecture to make the functions more clear after the release.

projects/step-frontend/src/lib/modules/timeseries/performance-view/performance-view.component.ts
dvladir: Instead of having the state arrays `initializationTasks`, `updateTasks`, I would suggest to the following approach.
Methods like `createSummaryChart`, `crateByStatusChart` should return observables. For example:
```
createSummaryChart(request: FindBucketRequest): Observable<unknown> {
  return this.timeSeriesService.fetchBuckets(request)
     .pipe(tap(response) => {
        this.createChart(TsChartType.OVERVIEW, request, response);
    });
}
```
So no additional subjects will require. And return observable will be **cold** (autocompleted), because it is a http requset. (In comparison subjects are **hot** observables - not autocompleted)

After that change `createAllCharts` will look like:
```
createAllCharts(): void {
   const charts$ = [
      this.createSummaryChart(this.findRequest),
      this.createByStatusChart(this.findRequest),
      this.createByKeywordsCharts(this.findRequest)
   ];
   if (this.includeThreadGroupChart) {
     charts$.push(this.createThreadGroupsChart(this.findRequest));
   }
   forkJoin(charts$).subsribe(() => this.onInitiailizationComplete.emit());
}
```

dvladir (dismissed):
1. All subjects and observables should have a `$` suffix at the end. So it should be `tableInitizlied$` instead of `tableInitializedSubject`.
2. I don't like the approach where lots of subjects are created and stored in the arrays. And then arrays could be modified and cleared. Potential risk to receive the memory leak. In one of comments I post a suggestion, how the code could be reorganized to avoid subjects creation

neogucky (approved):

PR #194, opened 2022-10-12
SED-1061 is used by dialog modal

chillbirdo:
### Old vs New Modal
Update: a component, that creates a modal window and wraps content has been introduces.

I think this way Modal windows can be handled in an intuitive way by the developer and directly interact with the component it is used in:
image

It is also possible to insert custom buttons to it using `<div step-modal-window-buttons><button>myButton</button></div>`

Everything works fine, however, there are errors in the browser's console:
image

Also, these errors are forcing me to comment-out all `tap`s that operate on instance variables:
https://github.com/exense/step-frontend/pull/194/files#diff-c57d3eb0d703ada28c1fe0ae54c4e6d1b6da559f13390646d08a327698019405R37

In case we can not get rid of these errors, another - yet much uglier - approach would be this: https://medium.com/swlh/load-dynamic-components-inside-modal-window-in-angular-2-5a570633aa1e

### Tenant Info from OS
Update: it is now loaded from RootScope.

### TBD:
- [ ] check whether legacy controller's code can be deleted and if so, delete. For now I left it, waiting for your feedback.


projects/step-frontend/src/lib/modules/_common/components/is-used-by-list/is-used-by-list.component.ts
```diff
@@ -0,0 +1,90 @@
...
+    switchMap(() =>
+      this._referencesService.findReferences({
+        searchType: this.type as SearchType,
+        searchValue: this.id,
+      })
+    ),
+    tap((result) => {
+      this.emptyResults = result.length === 0;
+      if (!this.emptyResults) {
+        const firstResultHasProject = !!result[0].attributes!['project'];
```
neogucky, line 45: We should replace it with: 

```suggestion
        const firstResultHasProject = !!result[0].attributes?.['project'];
```

neogucky: I see we are using this attribute with guarantee in other places too so it doesn't make sense to replace it here. In general I don't like guaranteeing variables. Check if we can move the others to optional (?), if not this can stay.

chillbirdo: this code has been removed anyways due to loading of tenant from rootScope

neogucky:
@chillbirdo Regarding the legacy code, do you mean for the isUsedBy dialog? Please double check that it is not used and if no remove it. After that we can merge.

chillbirdo:
Will do so on Monday

PR #196, opened 2022-10-12
SED-1058 executionError.html

dvladir:


projects/step-frontend/src/lib/angularjs/partials/execution/executionError.html
```diff
@@ -1,4 +1,15 @@
...
+  <step-execution-errors
+    [(selected-error-distribution-toggle)]="selectedErrorDistirbutionToggle"
+    [count-by-error-msg]="countByErrorMsg"
+    [count-by-error-code]="countByErrorCode"
+    [error-distribution]="errorDistribution"
+    (search-step-by-error)="searchStepByError($event)"
+  >
+  </step-execution-errors>
+
+<!--
```
neogucky, line 12: Not a fan of commenting out code, why is this needed?

neogucky:
Is the old code used elsewhere or why is this migration task only adding new code but not removing old one?

projects/step-frontend/src/lib/angularjs/partials/execution/executionError.html
dvladir: I left it, as it was left in other templates too.
If you think that it should be removed. Lest remove

neogucky (approved):

PR #199, opened 2022-10-13
SED-1464 empty-time-buckets-not-properly-handled

david-stephan:


neogucky (approved):
Perfectly kept in ticket scope, thanks!

PR #201, opened 2022-10-14
Sed 1443 ts   missing legend for total throughput on  throughput  dashlet

lucianbaciu:


projects/step-frontend/src/lib/modules/timeseries/chart/tooltip-plugin.ts
```diff
@@ -15,8 +16,15 @@ interface Anchor {
...
 }
 
+interface TooltipPluginSettings {
+  yScaleUnit?: string; // string to append on the y axis values
+  zAxisLabel?: string;
+}
+
 export class TooltipPlugin {
-  public static getInstance(yScaleUnit?: string): uPlot.Plugin {
+  static readonly SUMMARY_BARS_COLOR = '#97a7d7';
```
neogucky, line 25: We should have somewhere a colours modell containing all colours used in the visualization so it is easy to override them if we need to support different colour schemes

neogucky: Probably nothing unique to this PR, just a general observation

lucianbaciu: for now I will move them in my time-series.config.ts file

projects/step-frontend/src/lib/modules/timeseries/chart/tooltip-plugin.ts
```diff
@@ -1,6 +1,7 @@
 //@ts-ignore
 import uPlot = require('uplot');
 import { PlacementFunction } from './placement-function';
+import { set } from 'husky';
```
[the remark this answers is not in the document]
lucianbaciu: fixed

PR #202, opened 2022-10-14
SED-1395-Use terminator$ approach in order to destroy subscriptions (…

lucianbaciu:
…per Vladimir suggestion)

### DIFF
--- a/projects/step-frontend/src/lib/modules/timeseries/performance-view/measurements/measurements-picker.component.ts
+++ b/projects/step-frontend/src/lib/modules/timeseries/performance-view/measurements/measurements-picker.component.ts
@@ -18,7 +18,7 @@ export class MeasurementsPickerComponent implements OnInit, OnDestroy {
   allSeriesChecked: boolean = true;
   activeKeywords = 0;
 
-  subscriptionsTerminator$ = new Subject<void>();
+  terminator$ = new Subject<void>();
 
   constructor(private executionPageService: TimeSeriesContextsFactory) {}
 
@@ -29,7 +29,7 @@ export class MeasurementsPickerComponent implements OnInit, OnDestroy {
     this.keywordsService = this.executionPageService.getContext(this.contextId).getKeywordsContext();
     this.keywordsService
       .onKeywordsUpdated()
-      .pipe(takeUntil(this.subscriptionsTerminator$))
+      .pipe(takeUntil(this.terminator$))
       .subscribe((keywords) => {
         this.keywords = keywords;
         this.activeKeywords = 0;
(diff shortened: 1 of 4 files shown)


projects/step-frontend/src/lib/modules/timeseries/performance-view/measurements/measurements-picker.component.ts
```diff
@@ -75,8 +78,6 @@ export class MeasurementsPickerComponent implements OnInit, OnDestroy {
   };
 
   ngOnDestroy(): void {
-    if (this.subscriptions) {
-      this.subscriptions.unsubscribe();
-    }
+    this.subscriptionsTerminator$.next();
```
dvladir, line 81: It should be completed also

projects/step-frontend/src/lib/modules/timeseries/performance-view/measurements/measurements-picker.component.ts
```diff
@@ -18,7 +18,7 @@ export class MeasurementsPickerComponent implements OnInit, OnDestroy {
   allSeriesChecked: boolean = true;
   activeKeywords = 0;
 
-  subscriptions: Subscription = new Subscription();
+  subscriptionsTerminator$ = new Subject<void>();
```
dvladir, line 21: I would prefer to have a shorter name
just `terminator$` or `destroy$`

projects/step-frontend/src/lib/modules/timeseries/performance-view/time-selection/performance-view-time-selection.component.ts
```diff
@@ -157,8 +158,6 @@ export class PerformanceViewTimeSelectionComponent implements OnInit, OnDestroy
   }
 
   ngOnDestroy(): void {
-    if (this.subscriptions) {
-      this.subscriptions.unsubscribe();
-    }
+    this.subscriptionsTerminator$.next();
```
dvladir, line 161: Completion missing

dvladir (changes requested):
Check comments

projects/step-frontend/src/lib/modules/timeseries/performance-view/measurements/measurements-picker.component.ts
lucianbaciu: @dvladir I did the changes, but it seems to work without takeUntil. are you sure it is needed? maybe it's safer to complete it, but takeUntil will react when a .next() is called on the observable. Do I miss something?

dvladir: @lucianbaciu 
I played around a bit. Yes you're right `takeUntil` completes the source stream, when subject's `next` is invoked.
So `complete` inactivation not required for termination - I was wrong about it

But anyway completion of any subjects, that were declared in the component - is a good practice. I highly recommend to keep it.

dvladir (approved):

neogucky (approved):

PR #206, opened 2022-10-24
SED-1369-Replace trunc() with round()

lucianbaciu:


neogucky (commented):
Looks fine to me, just one question, it seems that we previously had avg values with decimal places and we are now changing to have none, is this the expected outcome?

david-stephan (approved):
as long as we deal with milliseconds, rounding to int is OK

PR #209, opened 2022-10-30
SED-1475 add metric selection for throughput chart

lucianbaciu:


projects/step-frontend/src/lib/modules/timeseries/execution-page/time-series-keywords.context.ts
```diff
@@ -14,12 +14,16 @@ export class TimeSeriesKeywordsContext {
...
 
-  private colorsPool: TimeseriesColorsPool;
+  private readonly colorsPool: TimeseriesColorsPool;
 
   constructor(colorsPool: TimeseriesColorsPool) {
     this.colorsPool = colorsPool;
   }
 
+  getColorsPool() {
+    return this.colorsPool;
```
dvladir, line 24: Remove this method, as field is already readonly

projects/step-frontend/src/lib/modules/timeseries/execution-page/time-series-keywords.context.ts
```diff
@@ -14,12 +14,16 @@ export class TimeSeriesKeywordsContext {
   private keywordToggleSubject = new Subject<KeywordSelection>();
   private selectAllSubject = new BehaviorSubject<boolean>(true);
 
-  private colorsPool: TimeseriesColorsPool;
+  private readonly colorsPool: TimeseriesColorsPool;
 
   constructor(colorsPool: TimeseriesColorsPool) {
     this.colorsPool = colorsPool;
   }
```
dvladir, line 21: From line 17 to 21 - it can be shorten as:
```
constructor(public readonly colorsPool: TimeseriesColorsPool) {
}
```
The public readonly field will be automatically created and will it take value from the class constructor

projects/step-frontend/src/lib/modules/timeseries/performance-view/measurements/measurements-picker.component.html
```diff
@@ -1,42 +1,31 @@
...
+      type="checkbox"
+      class="checkbox"
+      [(ngModel)]="allSeriesChecked"
+      (change)="onAllSeriesCheckboxClick($event)"
+      style="display: inline"
+    />
+    <span>All keywords</span>
+  </div>
+  <ng-container *ngFor="let item of keywords | keyvalue: valueAscOrder">
+    <ng-container *ngIf="!keywordSearchValue || item.key.toLowerCase().indexOf(keywordSearchValue) >= 0">
```
dvladir, line 16: It's not optimal.
Iteration has done over all entities, and condition is checked for every item during each change detection cycle.
This can be solved in the following way.

In the component create a form control
```
readonly keywordSearchValueCtrl: FormControl;

constructor(
  formBuilder: FormBuilder
){
  this.keywordSearchValueCtrl = formBuilder.control('');
}

```
Then bind it to input
```
<input [formControl]="keywordSearchValueCtrl" placeholder="Search" style="width: 110px; margin-bottom: 12px" />
```

In component create a rxjs stream of filtered items, that is based on controls'value:
```
readonly keywordItems$ = this.keywordSearchValueCtrl.valueChanges.pipe(
  startWith(''),
  distinctUntilChanged(),
  map((searchValue) => {
      // make a filtration here
  }) 
)
```
In template iterate over items in stream
```
<ng-container *ngFor="let item of (keywordItems$ | async)">
...
</ng-container>
```

dvladir (dismissed):
Check comments above

neogucky (commented):
Looks good, but I agree with Vladimirs points

projects/step-frontend/src/lib/modules/timeseries/performance-view/measurements/measurements-picker.component.html
lucianbaciu: Solved! keyvalue pipe is a problematic one since it's NOT pure. I created a new pipe which transform and filter

projects/step-frontend/src/lib/modules/timeseries/execution-page/time-series-keywords.context.ts
neogucky: @lucianbaciu we had a discussion about our coding conventions for setter / getter and decided to use public properties or the set/get syntax of TS. For this case we already have the exposed readonly property so we don't need anything else.

lucianbaciu: @neogucky alright. I will check for other getter/setters

neogucky (approved):

PR #210, opened 2022-10-30
SED-1463-Progress bar: increased opacity

lucianbaciu:


dvladir (dismissed):

projects/step-frontend/src/lib/modules/timeseries/performance-view/performance-view.component.html
```diff
@@ -9,7 +9,7 @@
       [settings]="settings"
     >
     </step-execution-time-selection>
-    <div style="height: 12px; opacity: 0.1">
+    <div style="height: 12px; opacity: 0.2">
```
neogucky, line 12: This should be performance-view.component.scss

neogucky (changes requested):
Please move styles to the scss file

neogucky (approved):

PR #211, opened 2022-10-30
SED-1484-Fix date and times fromatting

lucianbaciu:


dvladir (dismissed):

projects/step-frontend/src/lib/modules/timeseries/time-selection/time-range-picker.component.ts
```diff
@@ -140,12 +139,32 @@ export class TimeRangePicker implements OnInit {
...
+    let seconds = this.formatTimeValue(date.getSeconds());
+    let isoTime = `${hours}:${minutes}:${seconds}`;
+    return `${isoDate} ${isoTime}`;
+  }
+
+  formatTimeValue(value: number) {
+    if (value < 10) {
+      return '0' + value;
+    } else {
+      return value;
```
neogucky, line 166: might be nicer as:

`String(value).padStart(2, '0');`

neogucky: I would probably directly use this function where needed since at least for me padStart is clearer then "formatTimeValue" to understand what the function does.

neogucky (changes requested):
Looks good, but use padStart instead of formatTimeValue

projects/step-frontend/src/lib/modules/timeseries/time-selection/time-range-picker.component.ts
lucianbaciu: it's way better. I wasn't aware of this method.

neogucky (approved):

PR #213, opened 2022-11-01
SED-1062 repository testset selection

dvladir:


neogucky (changes requested):
Please select all testcases by default

dvladir:
> Please select all testcases by default

Implemented

PR #214, opened 2022-11-01
SED-1501 menu sidebar version container hidden in collapsed state

bojidaryovchev:
I also changed the margin-bottom calc to use 16px instead of 10px as the padding above is 16px, idk if this 10px was intentional tho, let me know if I should revert it

### DIFF
--- a/projects/step-core/styles/components/_step-sidebar.scss
+++ b/projects/step-core/styles/components/_step-sidebar.scss
@@ -222,7 +222,7 @@ $default-color: dc.default-colors();
 
     div#sidebar-categories-tabs {
       $helpSupportHeight: map.get($config, help-and-support-height);
-      margin-bottom: calc(#{$helpSupportHeight} + 10px);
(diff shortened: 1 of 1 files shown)


projects/step-core/styles/components/_step-sidebar.scss
```diff
@@ -403,9 +403,12 @@ $default-color: dc.default-colors();
         width: map.get($config, width-closed);
       }
 
+      #sidebar-categories-tabs {
+        margin-bottom: 16px;
+      }
```
neogucky, line 408: This doesn't have any effect and would mess up our margin if it did.

neogucky:
> I also changed the margin-bottom calc to use 16px instead of 10px as the padding above is 16px, idk if this 10px was intentional tho, let me know if I should revert it

Good catch. I guess it was set to 10 because it looks nice but I agreee we should have the same for top and bottom.

neogucky (approved):

PR #217, opened 2022-11-02
SED-1489 Update to Angular 14

dvladir:


neogucky (changes requested):
Everything is fine with the PR but we need to setup our build pipeline correctly before we do this merge.

neogucky (approved):

neogucky (approved):

PR #223, opened 2022-11-08
SED-1513 created a dialog to display STEP versions

bojidaryovchev:


dvladir:
@bojidaryovchev As other app's modals have buttons in the footer, I think it would be correct to add a close button to this new modal too.

dvladir:
@neogucky 
I am not sure the version's replacement works properly after it was moved from `index.html` to sidebar component and login page. On the deployed versions, I see the placeholders in the html layout.

dvladir (changes requested):
@bojidaryovchev Add the button to modal

@neogucky If version's replacement during the pipeline doesn't work properly, we need to fix it first, before merging this PR

neogucky:
> @neogucky I am not sure the version's replacement works properly after it was moved from `index.html` to sidebar component and login page. On the deployed versions, I see the placeholders in the html layout.

Actually we have some problem with the placeholders at the moment. In theory they should be replaced in all html and js files

neogucky (changes requested):
Looks very good, thank you! You couldn't know this but the version placeholders are replaced over all htmkl files in our build version, so you can just add them directly into the modal instad of adding them as parameters which makes the modal more streight forward.

projects/step-frontend/src/lib/modules/_common/components/versions-dialog/versions-dialog.component.html
```diff
@@ -0,0 +1,42 @@
...
+        <strong class="flex-1">Version</strong>
+      </div>
+
+      <div class="list-item">
+        <div class="flex-1">STEP</div>
+        <div class="flex-1">
+          <span style="display: none" id="frontend-os">{{ '${project.versionFEOS}' }}</span>
+          <span style="display: none" id="frontend-ee">{{ '${project.versionFEEE}' }}</span>
+        </div>
+      </div>
```
neogucky, line 17:
```suggestion
      <div class="list-item" id="frontend-os">
        <div class="flex-1">Step base</div>
        <div class="flex-1">
          <span style="display: none">{{ '${project.versionFEOS}' }}</span>
        </div>
      </div>
    <div class="list-item"  id="frontend-ee">
        <div class="flex-1">Step enterprise</div>
        <div class="flex-1">
          <span style="display: none">{{ '${project.versionFEEE}' }}</span>
        </div>
      </div>
```

neogucky: Please check code if you accept the suggestion, I just wrote it untested. The idea would be that we have two Step versions in Step EE and it is super helpful for us to know which ones are deployed. Makes it more clear to have them in their own row

projects/step-frontend/src/lib/modules/_common/components/versions-dialog/versions-dialog.component.ts
```diff
@@ -0,0 +1,55 @@
...
+    },
+    {
+      id: VersionDialogTabs.BE,
+      label: 'BackEnd',
+    },
+  ];
+  readonly VersionDialogTabs = VersionDialogTabs;
+
+  versionFEOS?: string;
+  versionFEEE?: string;
```
neogucky, line 29: These aren't used anymore, correct?

projects/step-frontend/src/lib/modules/_common/components/versions-dialog/versions-dialog.component.html
neogucky: And I think currently the version is not hidden when ${project.versionFEEE} is not replaced, correct?

neogucky (changes requested):
@bojidaryovchev Thanks, I left some comments! Can you also check if we can use this with the modal component?

projects/step-frontend/src/lib/modules/_common/components/versions-dialog/versions-dialog.component.ts
bojidaryovchev: nice catch

projects/step-frontend/src/lib/modules/_common/components/versions-dialog/versions-dialog.component.html
bojidaryovchev: well I am not really sure, I used the snipped you've suggested and this is what I see 
image

bojidaryovchev: however I am not really sure about those ids being attached to the list-item divs, are you sure this is how it should be? I can't really make an informed decision here

bojidaryovchev: image

neogucky: You are right, we need to also pull up the display:none into the list-item level. And we also need the logic to hide them when the placeholder is not replaced or empty. I will check tomorrow with DevOps so we can hopefully test it.

neogucky (approved):

PR #229, opened 2022-11-15
SED-1459 Migrate executionTree.html / reportTree

dvladir:


projects/step-core/src/lib/modules/tree/components/tree/tree.component.html
```diff
@@ -18,7 +18,7 @@
   >
     <button mat-button mat-icon-button disabled></button>
     <div class="node-content" (contextmenu)="openContextMenu($event, node.id)" cdkDragHandle>
-      <step-icon [name]="node._class | artefactIcon"></step-icon>
+      <step-icon [name]="node.icon" [class]="node.iconClassName"></step-icon>
```
neogucky, line 21: What is the class doing with the step-icon?

dvladir: For execution tree it is required to highlight the the icon with status color. To do it, I've added an ability to tree component that allows to customize icon's styles with additional css class

neogucky (approved):

PR #231, opened 2022-11-15
SED-1078 planName migration + structural improvements

bojidaryovchev:


projects/step-core/src/lib/components/plan-link/plan-link.component.html
```diff
@@ -0,0 +1,23 @@
...
+
+<ng-template #iconTemplate>
+  <step-icon name="file" matTooltip="Go to plan" matTooltipPosition="below"></step-icon>
+</ng-template>
+
+<ng-template #contentTemplate>
+  <ng-container *ngIf="!iconOnly; else iconTemplate">
+    <ng-container *ngTemplateOutlet="descriptionTemplate"></ng-container>
+  </ng-container>
+</ng-template>
```
dvladir, line 13: Can be simplified as
```
<ng-template #contentTemplate>
   <ng-container *ngIf="iconOnly; then iconTemplate; else descriptionTemplate">
   </ng-container>
</ng-template>
```

projects/step-core/src/lib/pipes/plan-name.pipe.ts
```diff
@@ -0,0 +1,11 @@
...
+import { Plan } from '../step-core.module';
+
+@Pipe({
+  name: 'planName',
+})
+export class PlanNamePipe implements PipeTransform {
+  transform(plan: Plan): string | undefined {
+    return plan.attributes?.['name'];
+  }
+}
```
dvladir, line 11: Is this pipe used somewhere?

dvladir (changes requested):
@bojidaryovchev  Pls, check comments above

projects/step-core/src/lib/pipes/plan-name.pipe.ts
bojidaryovchev: yes, I am using it in the EE version (there is a corresponding branch over there too)

neogucky (approved):

PR #238, opened 2022-11-22
SED-1056 Refactor Table progress.html

dvladir:


neogucky (changes requested):
It seems that the Execution functions aren’t working anymore (play button in the Execution sidebar)

dvladir:
@neogucky I have to restore `newShedulerTaskDialog.html` that was previously deleted. It was found out that it is still used, and we haven't migrated it yet

neogucky (approved):
Looks good now, thanks!

PR #239, opened 2022-11-22
SED-1079 notificationSubscribtionSelectionTable migration

bojidaryovchev:
- look aligned with the old version (padding)
- tooltips and such not visible inside old bootstrap modals (z-index)


projects/step-core/src/lib/modules/table/components/table/table.component.scss
```diff
@@ -15,8 +15,9 @@
     width: auto !important;
   }
 
-  ::ng-deep tr > td.mat-cell:not(:first-of-type):not(.mat-column-actions) {
-    padding: 0 8px;
+  ::ng-deep tr > td.mat-cell:not(:first-of-type):not(.mat-column-actions),
+  ::ng-deep tr > th.mat-header-cell:not(:first-of-type):not(.mat-column-actions) {
+    padding: 8px;
```
dvladir, line 20: Is this change required?
As far as I remember original line that was removed, was the part of the design fixes

neogucky (approved):
Looks fine, still we will review the padding: 8px (so top bottom are new) after merging.

PR #242, opened 2022-11-23
SED-1564 It is difficult to drag & drop into empty element

dvladir:


neogucky (changes requested):
Feature is deployed here https://stepos-sed-1564-tree-drag.exense.ch/

Doesn't work as expected for me:

General:
- I'm now not able at all to drag & drop something in a node that has no children

Design:
- Instead of making the target bold it would feel more intuitive to mark the element with a grey background and don't create a space above / beneath it. I added an image in the issue to demonstrate.

PR #243, opened 2022-11-24
SED-1521 disabling the scheduler

neogucky:


projects/step-core/src/lib/client/augmented/services/augmented-scheduler-service.ts
```diff
@@ -7,6 +7,8 @@ import { HttpClient } from '@angular/common/http';
 
 import { SchedulerService } from '../../generated';
 import { BaseHttpRequest } from '../../generated/core/BaseHttpRequest';
+import { map } from 'rxjs/operators';
+import { tap } from 'rxjs';
```
dvladir, line 11: No usage of these operators

projects/step-core/src/lib/client/augmented/services/augmented-scheduler-service.ts
```diff
@@ -26,4 +28,8 @@ export class AugmentedSchedulerService extends SchedulerService {
     //@ts-ignore
     return this._httpClient.post<any>('rest/scheduler/task/' + id + '/execute', null, { responseType: 'text' });
   }
+
+  public isSchedulerEnabled(): Observable<boolean> {
+    return this._httpClient.get<any>('rest/settings/scheduler_enabled');
```
dvladir, line 33: If you know that result will be boolean, it make sense to write invocation as `this._httpClient.get<boolean>`

dvladir (dismissed):
I've put some non critical remarks. But it's all find in common

projects/step-core/src/lib/client/augmented/services/augmented-scheduler-service.ts
neogucky, line 33:
```suggestion
    return this._httpClient.get<boolean>('rest/settings/scheduler_enabled');
```

projects/step-core/src/lib/client/augmented/services/augmented-scheduler-service.ts
neogucky, line 11:
```suggestion
```

dvladir (approved):

PR #257, opened 2022-12-06
SED-1604 Parameter table not reloaded after parameter edit

dvladir:


jeromecomte (approved):
@dvladir I Approve the changes. As I see the reloadTable should be registered for each CustomComponent serving as edit link. Could you check if this is registered for all other components?

PR #259, opened 2022-12-12
SED-1596 Filtering on the entity selection is broken

dvladir:


projects/step-frontend/src/lib/angularjs/js/ntables.js
```diff
@@ -378,7 +380,8 @@ angular
                   if (scope.serverSideParameters) {
                     data.params = scope.serverSideParameters;
                   }
-                  return JSON.stringify(tableLegacyUtils.transformRequestLegacy2New(data));
+                  lastServerSideRequest = tableLegacyUtils.transformRequestLegacy2New(data);
```
jeromecomte, line 383: @dvladir Do we still have legacy tables?

projects/step-frontend/src/lib/angularjs/js/app.js
```diff
@@ -959,27 +959,9 @@ angular
...
-              searchQuery['columns[' + i + '][search][regex]'] = column.search.regex;
-              searchQuery['order[0][column]'] = (order && order.column) || 0;
-              searchQuery['order[0][dir]'] = (order && order.dir) || 'asc';
-              searchQuery['draw'] = (oAjaxData && oAjaxData.draw) || 1;
-              searchQuery['start'] = (oAjaxData && oAjaxData.start) || 0;
-              searchQuery['length'] = (oAjaxData && oAjaxData.length) || 10;
-            }
-            return searchQuery;
+          $scope.handle.getFilters = function () {
+            const lastRequest = $scope.tableHandle.getLastServerSideRequest();
```
jeromecomte, line 963: @dvladir I assume this works differently with the new tables i.e. that the filters are rebuilt each time

jeromecomte (approved):

projects/step-frontend/src/lib/angularjs/js/ntables.js
dvladir: @jeromecomte Yes. 
For example assign entities modal, that belongs to angularjs part uses legacy table

projects/step-frontend/src/lib/angularjs/js/app.js
dvladir: @jeromecomte 
As assign entities is based on legacy tables, I have to modify the table's logic to store the last request made by table and then extract filters from it, when it is required.

In case of new tables, we already have the similar functionality that allows to retrieve the request object. But you're right it works a bit differently - table stores filters and search values for different columns. By invoking table's method `getTableFilterRequest` - it builds the request object, based on stored values.

PR #267, opened 2022-12-20
SED-1556 moved the modal-window.component and all its related users to the step-core.module and implemented the dynamic-textfield, expression-input and select-plan components

bojidaryovchev:


projects/step-core/src/lib/components/dynamic-textfield/dynamic-textfield.component.scss
```diff
@@ -0,0 +1,59 @@
+.enterprise-theme {
```
dvladir, line 1: Core components shouldn't have any mentions about enterprise.

projects/step-core/src/lib/components/dynamic-textfield/dynamic-textfield.component.ts
```diff
@@ -0,0 +1,96 @@
...
+      useExisting: forwardRef(() => DynamicTextfieldComponent),
+      multi: true,
+    },
+  ],
+})
+export class DynamicTextfieldComponent {
+  @Input() label?: string;
+  @Input() tooltip?: string;
+  @Input() invalid: boolean = false;
+  @Input() required: boolean = false;
```
dvladir, line 26: If component implements `ControlValueAccessor` it's validity state should be determined by `ng-invalid` class, that is automatically set by assigned Control in case of it's invalidity.

At the same time Control's validity/invalidity behavior should be setup by form's validators and not by additional props.

projects/step-core/src/lib/components/expression-input/expression-input.component.scss
```diff
@@ -0,0 +1,54 @@
+.enterprise-theme {
```
dvladir, line 1: Core components shouldn't have any mentions about enterprise.

projects/step-core/src/lib/components/expression-input/expression-input.component.ts
```diff
@@ -0,0 +1,63 @@
...
+      useExisting: forwardRef(() => ExpressionInputComponent),
+      multi: true,
+    },
+  ],
+})
+export class ExpressionInputComponent implements ControlValueAccessor {
+  @Input() label?: string;
+  @Input() tooltip?: string;
+  @Input() invalid: boolean = false;
+  @Input() required: boolean = false;
```
dvladir, line 25: Same as above. If we are using angular's forms, we should rely on validation mechanism provided by framework.
Otherwise it creates a possibility of contradictions like:
- Form control is valid and non required, but input flags are set is true
- Vice versa. Form control is required and invalid, but input flags are set is false

projects/step-core/src/lib/components/function-link/function-link.component.ts
```diff
@@ -13,7 +14,7 @@ export class FunctionLinkComponent implements CustomComponent {
   @Input() context?: KeywordFunction;
 
   readonly noLink$ = (this._customColumnOptions?.options$ || of([])).pipe(
-    map((options) => options.includes('noEditorLink'))
+    map((options: any) => options.includes('noEditorLink'))
```
dvladir, line 17: Remove any. It's redundant here. Typescript automatically intercepts the correct type from source observable

projects/step-core/src/lib/components/select-plan/select-plan.component.scss
```diff
@@ -0,0 +1,20 @@
+.enterprise-theme {
```
dvladir, line 1: Core components shouldn't have any mentions about enterprise.

dvladir (changes requested):

neogucky (changes requested):
There seems to be a circular dependency: https://build.exense.ch/#/root/executions/63a2e1dcfd5f6139bc1142e1?tenant=build

Entry point @exense/step-core has a circular dependency on itself.

projects/step-core/src/lib/components/function-link/function-link.component.ts
bojidaryovchev: I was getting an error about this, now its gone, weird

dvladir (approved):

neogucky (approved):

PR #273, opened 2023-01-04
SED-1050 operations list migration

bojidaryovchev:


projects/step-frontend/src/lib/modules/operations/components/operations-list/operations-list.component.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  triggerRefresh(): void {
+    this.getOperations$.next();
+  }
+
+  private getThreadId(operationDetails: OperationDetails): string | undefined {
+    const tid = operationDetails.operation?.tid;
+    const threadId = tid ?? '';
+
+    return `${threadId}`;
```
neogucky, line 98: I don't get this, why not

return threadId? 

Also the return value "undefined" shouldn't be possible with this implementation

neogucky (changes requested):
Small question, see comments

neogucky (approved):

PR #280, opened 2023-01-09
Sed 1570 ts   implement filter bar with multiselect/free text/number/date items

lucianbaciu:


projects/step-core/src/lib/modules/basics/components/date-filter/date-filter.component.html.orig
```diff
@@ -0,0 +1,15 @@
+<<<<<<< HEAD
```
bojidaryovchev, line 1: .orig. files should not be included in the code

projects/step-frontend/src/lib/modules/execution/components/execution-progress/execution-progress.component.scss
```diff
@@ -1,5 +1,5 @@
 as-split {
-  height: 80vh;
+  height: calc(100vh - 134px);
```
bojidaryovchev, line 2: use flexbox, it is possible to achieve the same with flexbox, as-split should have flex 1

projects/step-frontend/src/lib/modules/timeseries/chart/time-series-chart.component.ts
```diff
@@ -164,7 +164,24 @@ export class TimeSeriesChartComponent implements OnInit, AfterViewInit, OnChange
         },
         ...settings.series,
       ],
-      hooks: {},
+      hooks: {
+        // init: [
```
bojidaryovchev, line 168: if something will be commented out and preserved, it would be better to provide a descriptive comment as to why it was not removed in the first place

projects/step-frontend/src/lib/modules/timeseries/chart/time-series-chart.component.ts
```diff
@@ -234,20 +251,20 @@ export class TimeSeriesChartComponent implements OnInit, AfterViewInit, OnChange
    * Switches the state of the specified series.
    */
   toggleSeries(id: string): void {
-    const index = this.seriesIndexesByIds[id];
```
bojidaryovchev, line 237: I dont see these variables being reassigned anywhere, however this is the sort of thing that falls into the personal preference category, but I would still argue that this only contributes to more changes without providing actual value

projects/step-frontend/src/lib/modules/timeseries/dashboard/grouping/ts-grouping.component.html
```diff
@@ -0,0 +1,29 @@
...
+  <button
+    mat-menu-item
+    *ngFor="let grouping of groupingOptions"
+    (click)="selectedGrouping = grouping; emitGroupingChange()"
+  >
+    {{ grouping.label }}
+  </button>
+  <div mat-menu-item [matMenuTriggerFor]="customGroupingMenu">Other</div>
+  <mat-menu #customGroupingMenu="matMenu">
+    <div (click)="$event.stopPropagation()" style="padding: 12px">
```
bojidaryovchev, line 14: please avoid inline styles

projects/step-frontend/src/lib/modules/timeseries/dashboard/grouping/ts-grouping.component.html
```diff
@@ -0,0 +1,29 @@
...
+      <mat-form-field class="w-100">
+        <input
+          type="text"
+          matInput
+          autocomplete="off"
+          placeholder="Fields (comma separated)"
+          [(ngModel)]="customGroupingString"
+        />
+      </mat-form-field>
+      <div>
```
bojidaryovchev, line 24: do we really need an empty div here?

projects/step-frontend/src/lib/modules/timeseries/dashboard/time-series-dashboard.component.ts
```diff
@@ -0,0 +1,135 @@
...
+  readonly ONE_HOUR_MS = 3600 * 1000;
+
+  @Input() settings!: TimeSeriesDashboardSettings;
+  @ViewChild(PerformanceViewComponent) performanceView!: PerformanceViewComponent;
+
+  timeRangeSelection: TimeRangePickerSelection = { type: RangeSelectionType.FULL };
+  context!: TimeSeriesContext;
+  performanceViewSettings: PerformanceViewSettings | undefined;
+  updateSubscription = new Subscription();
+  update$: Observable<any> | undefined;
```
bojidaryovchev, line 30: any should be used as a fallback if no other type could be found, please double-check if that is the case

projects/step-frontend/src/lib/modules/timeseries/dashboard/time-series-dashboard.component.ts
```diff
@@ -0,0 +1,135 @@
...
+    { label: 'Last Hour', timeInMs: this.ONE_HOUR_MS },
+  ];
+
+  constructor(private contextsFactory: TimeSeriesContextsFactory) {}
+
+  ngAfterViewInit(): void {}
+
+  ngOnInit(): void {
+    if (!this.settings) {
+      throw new Error('Settings input must be set');
```
bojidaryovchev, line 49: is this a common practice in the project? I mean the exclamation mark is used to assert that settings will always exist, it is totally natural for the component to not work if it doesn't, I don't see a point of throwing an error ourselves explicitly

projects/step-frontend/src/lib/modules/timeseries/dashboard/time-series-dashboard.component.ts
```diff
@@ -0,0 +1,135 @@
...
+  timeRangeOptions: RelativeTimeSelection[] = [
+    { label: 'Last 1 Minute', timeInMs: this.ONE_HOUR_MS / 60 },
+    { label: 'Last 5 Minutes', timeInMs: this.ONE_HOUR_MS / 12 },
+    { label: 'Last 30 Minutes', timeInMs: this.ONE_HOUR_MS / 2 },
+    { label: 'Last Hour', timeInMs: this.ONE_HOUR_MS },
+  ];
+
+  constructor(private contextsFactory: TimeSeriesContextsFactory) {}
+
+  ngAfterViewInit(): void {}
```
bojidaryovchev, line 45: this appears to not be used

projects/step-frontend/src/lib/modules/timeseries/dashboard/time-series-dashboard.component.ts
```diff
@@ -0,0 +1,135 @@
...
+      baseFilters: this.settings.contextualFilters,
+      dynamicFilters: [],
+      grouping: ['name'],
+    };
+    this.context = this.contextsFactory.createContext(contextParams);
+    this.performanceViewSettings = this.settings;
+    this.subscribeForContextChange();
+    this.throttledRefreshTrigger$
+      .pipe(
+        throttle(() => this.context.inProgressChange().pipe(filter((x) => !x))),
```
bojidaryovchev, line 63: quality code teaches us to avoid having variables named like x and so forth, you could simply call it isProgressChange

projects/step-frontend/src/lib/modules/timeseries/dashboard/time-series-dashboard.component.ts
```diff
@@ -0,0 +1,135 @@
...
+      })
+      .subscribe();
+  }
+
+  subscribeForContextChange(): void {
+    merge(this.context.onFiltersChange(), this.context.onGroupingChange())
+      .pipe(takeUntil(this.terminator$))
+      .subscribe(() => {
+        this.updateSubscription?.unsubscribe();
+        // this.throttledRefreshTrigger$.next(null);
```
bojidaryovchev, line 86: it is good to remove unused comments that provide no value

projects/step-frontend/src/lib/modules/execution/components/execution-progress/execution-progress.component.scss
dvladir: @bojidaryovchev It's required to specify the concrete height, so angular-split could draw dragging bar properly

projects/step-frontend/src/lib/modules/timeseries/chart/time-series-chart.component.ts
lucianbaciu: agree

projects/step-frontend/src/lib/modules/timeseries/dashboard/time-series-dashboard.component.ts
lucianbaciu: agree

projects/step-frontend/src/lib/modules/timeseries/dashboard/grouping/ts-grouping.component.html
lucianbaciu: Personally, I prefer to have inline style for situations like this, where there is a single element which needs a small padding or margin. Several people online sustain this approach, so I don't think it is definitely something bad. I see this thing easier to follow, less code, less classes, faster to render. I think it is more of a preference thing though, but if you are all against it, I will definitely align. Just le me know! @neogucky @dvladir  @bojidaryovchev

projects/step-frontend/src/lib/modules/execution/components/execution-progress/execution-progress.component.scss
lucianbaciu: @bojidaryovchev it was my first instinct too. I think theoretically it is possible, but the entire page should be restyled in order to at least try it.

projects/step-frontend/src/lib/modules/timeseries/dashboard/time-series-dashboard.component.ts
bojidaryovchev: I don't this it's a bad idea per se, all I am saying is that I have seen other components and none of them used such an approach, and it is really important for the code to follow the same rules and conventions, otherwise it gets really messed up

bojidaryovchev: this is something that should be first discussed with the team and agreed upon as a common approach to follow, and then applied everywhere (consistency), and also be included in the code style guide

PR #281, opened 2023-01-10
SED-1637 Move leftover admin logic to EE

dvladir:


projects/step-core/src/lib/modules/basics/services/auth.service.ts
```diff
@@ -72,8 +72,8 @@ export class AuthService implements OnDestroy {
     this.setContext(context);
   }
 
-  login(credentials: CredentialsDto): Observable<unknown> {
-    return this._http.post('rest/access/login', credentials).pipe(
+  login(username: string, password: string): Observable<unknown> {
+    return this._loginService.login(username, password).pipe(
```
neogucky, line 76: @dvladir I just noticed this here, it might regards the original branch (SED-1637): 
Didn't we move the authentication completely away from OS? In this case I would also expect the auth service to be either moved to EE or to have a clean way to register authHandlers from EE.

neogucky (approved):

bojidaryovchev (approved):

neogucky (approved):

PR #282, opened 2023-01-11
SED-1650 resource input migration

bojidaryovchev:
- implemented functionality for upload with progress in the augmented-resources-service
- introduced the progress-bar.component
- introduced the upload-container.component
- did some styles improvements for inputs (for example with matSuffix)
- replaced the old resource input component across the app


projects/step-frontend/src/lib/modules/resources/components/resouce-input/resouce-input.component.html
```diff
@@ -0,0 +1,128 @@
...
+        <input
+          type="text"
+          readonly="readonly"
+          matInput
+          [ngModel]="resourceFilename"
+          matTooltip="Drag and drop your file here or enter the path to a file accessible from the controller"
+        />
+
+        <ng-container matSuffix>
+          <a href="rest/resources/{{ resourceId }}/content" mat-stroked-button matTooltip="Download this file">
```
dvladir, line 69: I think it make sense to avoid the endpoint usage directly in the template.
As it belongs to API I would create a method in AugmentedResourceService, which accept the resourceId and return the url.
Then retrieve this url, when resourceId is changed and bind it to anchor.

In this case things, which are related to API will be at one place.

bojidaryovchev: yeah that makes sense, I'll apply it

projects/step-core/src/lib/client/augmented/services/augmented-resources-service.ts
```diff
@@ -14,7 +16,96 @@ export class AugmentedResourcesService extends ResourcesService {
...
+    );
+
+    return {
+      progress$,
+      response$,
+    };
+  }
+
+  downloadResource(resourceId: string): void {
+    window.location.href = `/rest/resources/${resourceId}/content`;
```
dvladir, line 101: API service shouldn't change the location.
It should just return the url.

I would prefer to keep the anchor, so we could additionally setup the target, or do any further manipulations with url

dvladir (changes requested):

dvladir (approved):

PR #284, opened 2023-01-16
SED-1653 removing fixed context root

neogucky:


projects/step-core/src/lib/client/generated/services/PrivateApplicationService.ts
```diff
@@ -42,7 +42,7 @@ export class PrivateApplicationService {
   public getWebPlugins(): Observable<Array<AbstractWebPlugin>> {
     return this.httpRequest.request({
       method: 'GET',
-      url: '/app/plugins',
+      url: 'app/plugins',
```
dvladir, line 45: It is a generate service. It shouldn't be changed.

I guess change should be applied in `client/generated/core/OpenAPI` where rest base url is configured

dvladir (changes requested):

projects/step-core/src/lib/client/generated/services/PrivateApplicationService.ts
neogucky: You are correct, I should have marked this as draft since we are still using it to test some issues.

bojidaryovchev (approved):

dvladir (approved):

PR #287, opened 2023-01-17
SED-1082 Refactor Tables: searchResourceDialog.html

dvladir:


neogucky (dismissed):

projects/step-core/src/lib/components/modal-window/modal-window.component.ts
```diff
@@ -1,13 +1,39 @@
...
+  /**
+   * Sometimes it's require to show the migrated modal above the legacy one
+   * This input can be used to specify the z-index for .cdk-overlay-container when modal appears
+   * As soon as all modals will be migrated, this logic should be removed
+   */
+  @Input() zIndex?: number;
+  private cdkOverlayContainer?: HTMLElement;
+  constructor(private _elRef: ElementRef<HTMLElement>, public _dialogRef: MatDialogRef<unknown>) {}
+
+  ngAfterViewInit(): void {
```
bojidaryovchev, line 21: this solution looks nice, however it only solves the case with dialogs, and we actually have to solve this problem for all material things that make use of the overlay - an actual example is when we have a mat-select inside an old dialog, the menu with the options gets hidden behind the dialog..

a solution could be to create a directive where this logic is encapsulated and then add that directive to the providers of components that encounter such a case

bojidaryovchev (approved):

projects/step-core/src/lib/components/modal-window/modal-window.component.ts
dvladir: z-index assign logic was moved to a separate directive

PR #295, opened 2023-01-23
SED-1644 introduce versioning of entities (plan, parameter)

neogucky:


projects/step-core/src/lib/components/restore-dialog/restore-dialog.component.ts
```diff
@@ -0,0 +1,56 @@
...
+  constructor(
+    private _augmentedResourceService: AugmentedResourcesService,
+    private _matDialogRef: MatDialogRef<RestoreDialogComponent>,
+    private _filterConditionFactory: FilterConditionFactoryService,
+    @Inject(MAT_DIALOG_DATA) public plan: { version: string; history: Observable<History[]> }
+  ) {
+    this.dataSource = new TableLocalDataSource(plan.history, {
+      searchPredicates: {
+        updateTime: (element, searchValue) => {
+          console.log(element, searchValue);
```
bojidaryovchev, line 28: a forgotten console.log perhaps?

projects/step-core/src/lib/components/restore-dialog/restore-dialog.component.ts
```diff
@@ -0,0 +1,56 @@
...
+  styleUrls: ['./restore-dialog.component.scss'],
+})
+export class RestoreDialogComponent {
+  readonly dataSource: TableLocalDataSource<History>;
+
+  constructor(
+    private _augmentedResourceService: AugmentedResourcesService,
+    private _matDialogRef: MatDialogRef<RestoreDialogComponent>,
+    private _filterConditionFactory: FilterConditionFactoryService,
+    @Inject(MAT_DIALOG_DATA) public plan: { version: string; history: Observable<History[]> }
```
bojidaryovchev, line 23: I would personally prefer to extract an interface for the data, so I can then pass it when opening the dialog e.g. 
```
const matDialogConfig: MatDialogConfig = {
  data: {...}
};
const dialogRef = this._matDialog.open(DialogComponent, matDialogConfig);
```

projects/step-core/src/lib/components/restore-dialog/restore-dialog.component.ts
```diff
@@ -0,0 +1,56 @@
...
+    });
+  }
+
+  searchByDate(col: SearchColDirective, date?: DateTime): void {
+    const condition = this._filterConditionFactory.singleDateFilterCondition(date);
+    col.search(condition);
+  }
+
+  restore(history: History): void {
+    if (history.id) {
```
bojidaryovchev, line 48: a matter of preference I guess, but I would return as soon as possible in order to reduce indentation, e.g.
```
if (!history.id) {
  return;
}
```

bojidaryovchev: I generally try to follow this approach everywhere

projects/step-core/src/lib/components/restore-dialog/restore-dialog.component.ts
```diff
@@ -0,0 +1,56 @@
...
+      this._matDialogRef.close(history.id);
+    }
+  }
+
+  cancel(): void {
+    this._matDialogRef.close(undefined);
```
bojidaryovchev, line 54: also a matter of preference, but there is no actual need to be explicit here

bojidaryovchev (approved):

bojidaryovchev:
I tried to track some time for this review, however I don't seem to have access
image

bojidaryovchev:
I spent around 15 mins to review this, where do I track them? I guess here? https://exense.atlassian.net/browse/SED-1649

dvladir (dismissed):

projects/step-core/src/lib/components/restore-dialog/restore-dialog.component.ts
neogucky: I actually like to be explicit because closing with undefined will have a different behaviour than closing with a value

neogucky:
> I spent around 15 mins to review this, where do I track them? I guess here? https://exense.atlassian.net/browse/SED-1649

Yes correct!

projects/step-core/src/lib/components/restore-dialog/restore-dialog.component.ts
bojidaryovchev: what I meant is that if you  dont pass in anything it will also result in undefined, therefore I don't think its necessary.. anyway tho, up to you

bojidaryovchev (dismissed):

bojidaryovchev (approved):

PR #308, opened 2023-01-31
SED-1702 Update plan editor content-split

bojidaryovchev:
- introduced custom split components and applied them to the plan editor and plan tree


projects/step-core/src/lib/components/split-area/split-area.component.ts
```diff
@@ -0,0 +1,63 @@
...
+  styleUrls: ['./split-area.component.scss'],
+})
+export class SplitAreaComponent implements AfterViewInit {
+  @Input() sizeType?: SplitAreaSizeType;
+  @Input() size?: string;
+  @Input() padding?: string;
+
+  constructor(private _elementRef: ElementRef<HTMLElement>) {}
+
+  ngAfterViewInit(): void {
```
bojidaryovchev, line 17: earlier today I showed this to Vladimir and he said that maybe it would make sense for this logic to happen in ngOnChanges, however I would argue it would be better for it to be here, because the idea is to set the initial dimensions only, and when we use the components we should not really have a case where we set some size and then change it dynamically, besides such a case is not present at all in our project;

the idea is that we can say for the initial size to be pixels, percents, flex (flex-grow) and we can also omit providing a size, thus letting the content expand the split area as needed, so generally we should not have to change the size dynamically (the padding is not really part of the calculations so it is also irrelevant)

neogucky (dismissed):

neogucky (approved):

PR #313, opened 2023-02-04
SED-1728 Incorrect identation of the visual plan editor tree

dvladir:


bojidaryovchev (approved):

dvladir:
It was decided to implement it in a bit different way

PR #314, opened 2023-02-04
SED-1731 Visual plan editor - Components are not added as sequence to the tree

dvladir:


bojidaryovchev (dismissed):

projects/step-core/src/lib/modules/tree/components/tree-node/tree-node.component.scss
```diff
@@ -22,36 +23,28 @@ step-tree-node {
...
-        }
-      }
     }
 
     &.skipped {
       color: var.$muted;
     }
+
+    &.tree-node-selected {
+      font-weight: 600;
```
neogucky, line 43:
```suggestion
      font-weight: 500;
```

neogucky (approved):

PR #316, opened 2023-02-04
SED-1713 Treeview: F2 rename node does not update the "title"

dvladir:


projects/step-core/src/lib/components/keyword-name/keyword-name.component.ts
```diff
@@ -30,7 +42,9 @@ const KEYWORD_ATTRIBUTES_SCHEMA: DynamicFieldsSchema = {
   templateUrl: './keyword-name.component.html',
   styleUrls: ['./keyword-name.component.scss'],
 })
-export class KeywordNameComponent implements OnChanges {
+export class KeywordNameComponent implements OnChanges, OnInit, OnDestroy {
+  private terminator$ = new Subject<unknown>();
```
bojidaryovchev, line 46: Why is the terminator a ```Subject<unknown>``` and not a ```Subject<void>```, and then why is an object forwarded with .next({}) instead of just ```.next()```? We need to pass anything in the terminator, right?

bojidaryovchev (approved):

projects/step-core/src/lib/components/keyword-name/keyword-name.component.ts
neogucky: I guess this is really inconsistent in our code, we have <any>, <unknown> and <void>. Lets have a discussion in the weekly and fix it in the codebase after. Since it concerns many components anyway I will just merge this for now.

dvladir: @bojidaryovchev On the one hand yes, but on the other subjects and observables are created to emit data, so passing nothing to it - it looks illogical in my point of view.

dvladir: @bojidaryovchev @neogucky 
An update. RxJS documentation explicitly mention the case about `Subject<void>` 
https://rxjs.dev/guide/subject#void-subject
So Bojidar's remark is absolutely correct

PR #319, opened 2023-02-05
SED-1951 Time series filters a11y

bojidaryovchev:
- introduced the focusable.directive
- introduced the focusables.directive
- introduced the trap-focus.directive
- introduced the recursive-tab-index.directive


bojidaryovchev:
what you see here is achieved by using the keys Tab, Shift + Tab, Space, Enter

projects/step-core/src/lib/directives/recursive-tab-index.directive.ts
```diff
@@ -0,0 +1,63 @@
...
+    bfsFilter: <T>({ predicate, children }: { predicate?: (item: T) => boolean; children: (item: T) => T[] }) => T[];
+  }
+}
+
+/**
+ * The time complexity of a Breadth-First Search (BFS) algorithm is O(|V| + |E|),
+ * where |V| is the number of vertices and |E| is the number of edges in the graph being traversed.
+ * The reason for this is that in the worst-case scenario, BFS needs to visit all vertices and all edges.
+ */
+Array.prototype.bfsFilter = function bfsFilter<T>({
```
dvladir, line 14: In my opinion this function shouldn't be included in Array's prototype.
It is used to walk through the tree structures. 
But array isn't a tree.

bojidaryovchev:
closing this to recreate it in a new branch

lucianbaciu:
filters have been drastically changed because of the last tasks. we should sync before merging new stuff

PR #339, opened 2023-02-13
SED-1774 Skipped node by inheritance are not displayed as disabled

bojidaryovchev:


neogucky (dismissed):

projects/step-core/src/lib/modules/tree/services/tree-state.service.ts
```diff
@@ -320,15 +336,26 @@ export class TreeStateService<T, N extends TreeNode> implements OnDestroy {
...
     if (!this.isPossibleToInsert(parentId, ...newNodes)) {
       return;
     }
 
+    const parent = this.findNodeById(parentId);
+
+    if (parent) {
+      newNodes.forEach((node) => {
+        node.isSkipped = parent.isSkipped;
+        (node as unknown as { originalArtefact: AbstractArtefact }).originalArtefact.skipNode!.value = parent.isSkipped;
```
neogucky, line 355: Why are we using two as here? Doesn't the second one override the first?

projects/step-core/src/lib/shared/utils.ts
```diff
@@ -37,3 +38,42 @@ export const setObjectFieldValue = (object: Record<string, unknown>, fieldPath:
     return (res as Record<string, unknown>)[fieldName];
   }, object);
 };
+
+/**
+ * The time complexity of a Breadth-First Search (BFS) algorithm is O(|V| + |E|),
+ * where |V| is the number of vertices and |E| is the number of edges in the graph being traversed.
+ * The reason for this is that in the worst-case scenario, BFS needs to visit all vertices and all edges.
+ */
+export const bfs = <T>({
```
neogucky, line 47:
```suggestion
export const bfSearch = <T>({
```

I would prefer so have a naming that makes it clear this is a search

projects/step-core/src/lib/modules/tree/services/tree-state.service.ts
dvladir, line 355: TreeStateService is abstract common thing. In theory it should know nothing about node's content.

Placing this logic here not a clean solution.
I would prefer to move somewhere else, where we have already know that we are working with AbstractArtefact

neogucky (approved):

dvladir (approved):

PR #342, opened 2023-02-13
SED-1773 Default value not set for boolean keyword input

bojidaryovchev:


projects/step-frontend/src/lib/modules/execution/components/execution-commands/execution-commands.component.ts
```diff
@@ -140,7 +140,7 @@ export class ExecutionCommandsComponent implements OnInit, OnChanges {
           includedNames: this.includedTestcases.list,
         };
       } else if (this.includedTestcases.by === 'all') {
-        delete artefactFilter;
+        (artefactFilter as any) = undefined;
```
bojidaryovchev, line 143: this wasnt compiling so I reverted it back to Vladimir's version

projects/step-core/src/lib/modules/dynamic-forms/components/dynamic-field/dynamic-field.component.ts
```diff
@@ -127,7 +127,19 @@ export class DynamicFieldComponent implements ControlValueAccessor, OnDestroy {
...
-    this.value = this.internalValue?.value;
+    let fallbackValue: string | number | boolean | undefined = undefined;
+
+    switch (this.fieldType) {
+      case DynamicFieldType.string:
+        fallbackValue = '';
+        break;
+
+      case DynamicFieldType.boolean:
+        fallbackValue = false;
```
neogucky, line 138: I would set this one to undefined, else booleans might be initialised with "false"

bojidaryovchev: not really, that is handled by the code below 
image

neogucky (approved):

PR #349, opened 2023-02-15
SED-1796 Restored scheduler settings to user settings page

neogucky:


bojidaryovchev:
@neogucky I think you accidentally added me as an assignee, instead of a reviewer 😆

bojidaryovchev (approved):

neogucky:
> @neogucky I think you accidentally added me as an assignee, instead of a reviewer 😆

True, thanks for the review!

PR #359, opened 2023-02-22
SED-1823 wrapping text for parameter value

neogucky:


projects/step-frontend/src/lib/modules/parameter/parameters-list/parameters-list.component.scss
```diff
@@ -2,6 +2,12 @@
 
 :host {
   @include core.table-container;
+
+  step-table {
+    ::ng-deep td.cdk-column-value {
```
bojidaryovchev, line 7: from what I know ::ng-deep should not be used as it is deprecated, the fact that it still works is not something we should be relying on in my opinion, the proper way to handle this would be to use ViewEncapsulation.None and then instead of :host you would use the component selector e.g. step-parameters-list { ... }

bojidaryovchev (approved):

projects/step-frontend/src/lib/modules/parameter/parameters-list/parameters-list.component.scss
neogucky: I'm not sure, ViewEncapsulation.None is quite dangerous. In my opinion ng-deep inside of :host is the better approach. I guess they only deprecate it so people double check it is really encapsulated in :host.

PR #365, opened 2023-02-23
SED-1834 Only start new request when last request finished

neogucky:


projects/step-core/src/lib/modules/table/shared/table-remote-data-source.ts
```diff
@@ -110,12 +110,16 @@ const convertTableRequest = (req: TableRequestInternal): TableRequestData => {
 export class TableRemoteDataSource<T> implements TableDataSource<T> {
   private _terminator$ = new Subject<any>();
   private _inProgress$ = new BehaviorSubject<boolean>(false);
+  private _requestLock$ = new BehaviorSubject<boolean>(false);
```
bojidaryovchev, line 113: this subject does not seem to be completed in ngOnDestroy

bojidaryovchev (dismissed):

projects/step-core/src/lib/modules/table/shared/table-remote-data-source.ts
neogucky: Good point, ngOnDestroy is commented out due to an issue but I added it there so we do not forget when commenting in again

PR #369, opened 2023-02-24
SED-1817-Labels changed

lucianbaciu:


neogucky (approved):
Thanks for making the dedicated PR, so much easier! :)

PR #373, opened 2023-03-03
SED-1829 Navigating back and forward doesn't work properly on all views

bojidaryovchev:


projects/step-frontend/src/lib/modules/_common/components/sidebar/sidebar.component.ts
```diff
@@ -99,21 +99,22 @@ export class SidebarComponent implements AfterViewInit, OnDestroy {
         this._authService.gotoDefaultPage();
         break;
       default:
-        if (this._ajsLocation.path().includes('/root/' + viewId)) {
```
bojidaryovchev, line 102: not sure what has this been used for, but I didn't really see an issue removing it

neogucky: We still require this. It reloads the current page, i.e. when you are currently on plans/list and click plans in the menu without this it will not reload "plans/list" since no URL change is triggered.

neogucky (changes requested):
Thanks for the PR, I like your solution since it simplifies the code. Two issues and one bonus task:

- we need the removed part in order to reload active page (I see a nicer solution but lets not waste time on the routing)
- "plans" menu item is currently not highlighted when active
+ The same issue will also exist for executions/list for me it makes sense to just also solve it as part of this ticket

bojidaryovchev:
I cannot solve this, closing the PR

PR #379, opened 2023-03-14
SED-1793 Execution tree not refreshed properly

dvladir:


projects/step-frontend/src/lib/angularjs/partials/execution/executionTree.html
```diff
@@ -6,8 +6,11 @@
     </div>
     <div class="panel-body" ng-show="isShowPanel('reportTree')">
       <div>
+<!--
+        Not used
         <step-execution-tree [node-id]="eid" [handle]="reportTreeHandle">
         </step-execution-tree>
+-->
```
neogucky, line 13: Probably we should remove the whole file? I see we have the complete old tab-bar still in our code but probably don't need them anymore, correct?

neogucky (approved):

projects/step-frontend/src/lib/angularjs/partials/execution/executionTree.html
dvladir: @neogucky Correct.
We still have legacy html/js in the project related to the execution page, that not used anymore.
But probably it make sense to create a separate ticket to cleanup those files.

bojidaryovchev (approved):

PR #381, opened 2023-03-15
SED-1859 user preferences not working

neogucky:


projects/step-core/src/lib/modules/table/services/items-per-page.service.ts
```diff
@@ -0,0 +1,34 @@
...
+  private defaultItemsPerPage: Array<number> = [10, 25, 50, 100];
+  private customItemsPerPage?: number | boolean;
+
+  constructor(private _userService: UserService) {}
+
+  public getItemsPerPage(loadedUserPreferences: Function): any {
+    if (this.customItemsPerPage === undefined) {
+      this._userService.getPreferences().subscribe((preferences) => {
+        if (preferences?.preferences?.['tables_itemsperpage']) {
+          this.customItemsPerPage = Number(preferences.preferences['tables_itemsperpage']);
```
bojidaryovchev, line 17: the expression ```Number(preferences.preferences['tables_itemsperpage'])``` exists twice, it would be nice to have it as a constant

projects/step-core/src/lib/modules/table/services/items-per-page.service.ts
```diff
@@ -0,0 +1,34 @@
...
+@Injectable({
+  providedIn: 'root',
+})
+export class ItemsPerPageService {
+  private defaultItemsPerPage: Array<number> = [10, 25, 50, 100];
+  private customItemsPerPage?: number | boolean;
+
+  constructor(private _userService: UserService) {}
+
+  public getItemsPerPage(loadedUserPreferences: Function): any {
```
bojidaryovchev, line 13: I think the type ```Function``` is too abstract, it would be better to define it as ```(itemsPerPage: number) => void``` imo

bojidaryovchev: also the return type seems to be ```number``` at a first glance, perhaps its ```any``` for a reason?

projects/step-core/src/lib/modules/table/services/items-per-page.service.ts
```diff
@@ -0,0 +1,34 @@
...
+    if (this.customItemsPerPage === undefined) {
+      this._userService.getPreferences().subscribe((preferences) => {
+        if (preferences?.preferences?.['tables_itemsperpage']) {
+          this.customItemsPerPage = Number(preferences.preferences['tables_itemsperpage']);
+          loadedUserPreferences(Number(preferences.preferences['tables_itemsperpage']));
+        }
+      });
+    }
+    const itemsPerPage = this.defaultItemsPerPage;
```
bojidaryovchev, line 22: you are assigning ```itemsPerPage``` without changing the reference (```this.defaultItemsPerPage.slice()``` or ```[...defaultItemsPerPage]```) so that later when you do ```.unshift(this.customItemsPerPage)``` it would also mutate ```this.defaultItemsPerPage```

projects/step-core/src/lib/modules/table/services/items-per-page.service.ts
```diff
@@ -0,0 +1,34 @@
+import { Injectable } from '@angular/core';
+import { UserService } from '../../../client/generated';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class ItemsPerPageService {
+  private defaultItemsPerPage: Array<number> = [10, 25, 50, 100];
```
bojidaryovchev, line 8: perhaps it makes sense to denote it as ```readonly``` as well, since its name suggests that it should not be changed

projects/step-core/src/lib/modules/table/components/table/table.component.ts
```diff
@@ -113,6 +113,8 @@ export class TableComponent<T> implements AfterViewInit, OnChanges, OnDestroy, T
 
   searchColumns: SearchColumn[] = [];
 
+  pageSizeOptions: Array<number>;
```
bojidaryovchev, line 116: this is something I believe we should actually discuss because it raises a valid point - as I see, you don't have ? or ! after ```pageSizeOptions```, and the reason you are not getting a compilation error is because it is assigned in the constructor; I think we should consider whether we should start making more use of our constructors for code that does not directly relate to the component lifecycle

bojidaryovchev: if we talk about components, code that can be put in constructors can also be seamlessly put in ```ngOnInit``` (at least most of times); if we talk about services, they don't have lifecycle hooks (except for ```ngOnDestroy``` if they are provided at the component level), so the code goes in the constructor; there is one more alternative - code that would go into a service's constructor could be extracted into a method, calling which would act as a constructor, but the downside is that in order to assign ```readonly``` properties in such a function we would have to use the ```(this as { foo: boolean }).foo = true``` approach, which we could save ourselves from if we just keep those initializations in the constructor (as it is totally fine to initialize ```readonly``` properties in the constructor)

bojidaryovchev: I will create a card on Trello for these aforementioned points and I will refer to this PR's discussion

bojidaryovchev (approved):

projects/step-core/src/lib/modules/table/services/items-per-page.service.ts
neogucky, line 8:
```suggestion
  readonly defaultItemsPerPage: ReadonlyArray<number> = [10, 25, 50, 100];
```

projects/step-core/src/lib/modules/table/services/items-per-page.service.ts
neogucky: The return type is actually Array<number>, but good point.

projects/step-core/src/lib/modules/table/services/items-per-page.service.ts
neogucky:
```suggestion
    const itemsPerPage = [...this.defaultItemsPerPage];
```

projects/step-core/src/lib/modules/table/services/items-per-page.service.ts
```diff
@@ -0,0 +1,34 @@
...
+  private customItemsPerPage?: number | boolean;
+
+  constructor(private _userService: UserService) {}
+
+  public getItemsPerPage(loadedUserPreferences: Function): any {
+    if (this.customItemsPerPage === undefined) {
+      this._userService.getPreferences().subscribe((preferences) => {
+        if (preferences?.preferences?.['tables_itemsperpage']) {
```
neogucky, line 16:
```suggestion
        const tables_itemsperpage = preferences?.preferences?.['tables_itemsperpage'];
        if (tables_itemsperpage) {
```

projects/step-core/src/lib/modules/table/services/items-per-page.service.ts
neogucky:
```suggestion
          this.customItemsPerPage = Number(tables_itemsperpage);
```

projects/step-core/src/lib/modules/table/services/items-per-page.service.ts
```diff
@@ -0,0 +1,34 @@
...
+          this.customItemsPerPage = Number(preferences.preferences['tables_itemsperpage']);
+          loadedUserPreferences(Number(preferences.preferences['tables_itemsperpage']));
```
neogucky, line 18:
```suggestion
          loadedUserPreferences(Number(tables_itemsperpage));
```

projects/step-core/src/lib/modules/table/services/items-per-page.service.ts
david-stephan:
```suggestion
          this.customItemsPerPage = Number(preferences.preferences['tables_itemsperpage']);
```

bojidaryovchev (approved):

PR #389, opened 2023-03-20
SED-1721 Make warning more prominent when you're about to delete ALL items in a collection

dvladir:


neogucky (changes requested):
I like it, but request a small change:

Instead of highlighting "delete" in the preview body, I would highlight the number of elements affected. I would also replace in the all selected case the word "the" with "all":

"Do you want to delete **all 2 selected item(s)**?"

neogucky (approved):

PR #393, opened 2023-03-23
SED-1911 FE Configurable menu depending on role

dvladir:


projects/step-frontend/src/lib/modules/_common/components/sidebar/sidebar.component.ts
```diff
@@ -21,19 +23,24 @@ import {
...
 })
-export class SidebarComponent implements AfterViewInit, OnDestroy {
+export class SidebarComponent implements OnInit, OnDestroy {
   @ViewChildren('mainMenuCheckBox') mainMenuCheckBoxes?: QueryList<ElementRef>;
 
   sideBarOpen: boolean = true;
-
+  private terminator$ = new Subject<void>();
   private locationStateSubscription: any;
+  readonly _menuItems$ = inject(MENU_ITEMS).pipe(takeUntil(this.terminator$));
```
bojidaryovchev, line 41: why is this prefixed with an underscore? and why is inject used instead of the injection to happen in the constructor? I think I actually tried this at some point and it was not even working, but even if it does work, I would say not to apply it until a migration to a newer Angular version has begun

bojidaryovchev (dismissed):

projects/step-frontend/src/lib/modules/_common/components/sidebar/sidebar.component.ts
dvladir: @bojidaryovchev `inject` is available since v14. The project has been moved to v14 since the end of November

bojidaryovchev (approved):

PR #394, opened 2023-03-23
SED-1895 Plan editor: support copy/pasting between tabs

bojidaryovchev:


projects/step-core/src/lib/shared/storage.token.ts
```diff
@@ -0,0 +1,6 @@
+import { InjectionToken } from '@angular/core';
+
+export const STORAGE = new InjectionToken<Storage>('storage', {
+  providedIn: 'root',
+  factory: () => localStorage,
```
dvladir, line 5: @neogucky @bojidaryovchev 
Do we need to persist copied nodes after the user closed browser?
Probably `sessionStorage` will be more suitable for this case. It will be automatically cleared on browser's close

dvladir (approved):

dvladir (changes requested):
PR is fine, but target branch is wrong. Task belongs to 3.22

projects/step-core/src/lib/shared/storage.token.ts
bojidaryovchev: @dvladir sessionStorage does not support cross-tab communication, localStorage or indexedDB seem to be the most viable options, and since currently we do not have anything large enough to make sense to store it in indexedDB (perhaps we could think about store functionality which would make use of a storage adapter and then we can create one for indexedDB and use it to implement caching in certain places)

bojidaryovchev: I could however implement destruction functionality to clear out persistence tokens, I would have to introduce a convention for their naming in order to be able to efficiently clear them

dvladir (approved):

dvladir (approved):

PR #401, opened 2023-03-29
SED-1866 Implement our own component to replace mat form fields

dvladir:


projects/step-core/src/lib/modules/dynamic-forms/components/dynamic-field/dynamic-field.component.scss
```diff
@@ -2,16 +2,16 @@
 @use 'projects/step-core/styles/core-mixins' as core;
 
 step-dynamic-field {
-  .mat-button-wrapper step-icon {
-    @include core.step-icon-size(16px);
-  }
+  //.mat-button-wrapper step-icon {
```
bojidaryovchev, line 5: do we need this?

projects/step-core/src/lib/modules/dynamic-forms/components/expression-input/expression-input.component.html
```diff
@@ -1,34 +1,30 @@
-<mat-form-field floatLabel="always" class="step-view" [class.ng-invalid]="!!_ngControl.invalid || isParentInvalid">
-  <mat-label *ngIf="label">{{ label }}</mat-label>
+<step-form-field [class.ng-invalid]="!!_ngControl.invalid || isParentInvalid" [class.with-tooltip]="tooltip">
+  <step-label [attr.show-required-marker]="showRequiredAsterisk || null" *ngIf="label">{{ label }}</step-label>
```
bojidaryovchev, line 2: why do we need the ```|| null``` part

bojidaryovchev:
overall looks brilliant, I've just left some minor remarks

projects/step-core/src/lib/modules/dynamic-forms/components/expression-input/expression-input.component.html
dvladir: Without `null` attribute will remain, even if `showRequiredAsterisk` will be `false`.
In case of `null` angular will remove the attribute

projects/step-core/src/lib/modules/dynamic-forms/components/dynamic-field/dynamic-field.component.scss
dvladir: Nope. I forgot to remove

projects/step-core/src/lib/modules/dynamic-forms/components/expression-input/expression-input.component.html
bojidaryovchev: nice catch

bojidaryovchev (dismissed):

neogucky (dismissed):

neogucky (approved):

PR #404, opened 2023-03-31
SED-1863 Bulks support for remaining tables

dvladir:


bojidaryovchev (approved):

bojidaryovchev (approved):

projects/step-core/src/lib/modules/basics/services/item-by-id-cache.service.ts
```diff
@@ -0,0 +1,28 @@
+import { Injectable, OnDestroy } from '@angular/core';
+import { Observable, of, tap } from 'rxjs';
+
+@Injectable()
+export abstract class ItemByIdCacheService<T> implements OnDestroy {
+  private itemsCache = new Map<string, T>();
+
+  getItem(id: string): Observable<T> {
+    if (this.itemsCache.has(id)) {
+      return of(this.itemsCache.get(id)!);
```
bojidaryovchev, line 10: what you can do here instead of using .has() and then .get() with ! assertion, you can say ```const existingItem = this.itemsCache.get(id);``` and then ```if (existingItem) { return of(existingItem); }```

bojidaryovchev (approved):

projects/step-frontend/src/lib/modules/mask/components/services/generic-function-bulk-operations-invoke.service.service.ts
```diff
@@ -0,0 +1,22 @@
...
+import {
+  AsyncTaskStatus,
+  AugmentedKeywordsService,
+  BulkOperationsInvokeService,
+  TableBulkOperationRequest,
+} from '@exense/step-core';
+import { Observable } from 'rxjs';
+
+@Injectable()
+export class GenericFunctionBulkOperationsInvokeServiceService extends BulkOperationsInvokeService<string> {
```
bojidaryovchev, line 11: maybe ```GenericFunctionBulkOperationsInvokeService``` instead of ```GenericFunctionBulkOperationsInvokeServiceService```

bojidaryovchev (dismissed):

neogucky (approved):

projects/step-core/src/lib/modules/basics/services/repository-parameters-storage.service.ts
```diff
@@ -0,0 +1,24 @@
...
+import { Observable, tap } from 'rxjs';
+import { PrivateApplicationService } from '../../../client/step-client-module';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class RepositoryParametersStorageService {
+  private _api = inject(PrivateApplicationService);
+
+  private parametersInternal: ReadonlyArray<string> = [];
```
bojidaryovchev, line 11: you can also say ```= [] as const;```

projects/step-core/src/lib/modules/table/shared/basic-filter-condition.ts
```diff
@@ -0,0 +1,15 @@
...
+export class BasicFilterCondition extends FilterCondition {
+  constructor(private filters: TableRequestFilter[]) {
+    super();
+  }
+  override toRequestFilter(field: string): Array<TableRequestFilter | undefined> {
+    return this.filters;
+  }
+
+  override isEmpty(): boolean {
+    return !this.filters || this.filters.length === 0;
```
bojidaryovchev, line 13: is there a chance for ```this.filters``` to not exist? I mean the way it is declared, it should always be defined

projects/step-core/src/lib/modules/table/shared/parameters-filter-condition.ts
```diff
@@ -0,0 +1,36 @@
...
+  constructor(private availableParameters: ReadonlyArray<string>, private searchValue?: string) {
+    super();
+  }
+
+  override isEmpty(): boolean {
+    return !this.searchValue;
+  }
+
+  override toRequestFilter(field: string): Array<TableRequestFilter | undefined> {
+    if (!this.searchValue || this.availableParameters.length === 0) {
```
bojidaryovchev, line 15: I know this ```.length === 0``` is considered more explicit, but JavaScript specifically gives us the ability to work with truthy and falsy values in order to simplify the code, why not just write ```!this.availableParameters.length```

bojidaryovchev (dismissed):

projects/step-core/src/lib/modules/basics/services/repository-parameters-storage.service.ts
bojidaryovchev: also I guess it makes more sense to say ```private parametersInternal: readonly string[] = []``` because when we are defining arrays we say T[] and not Array<T>

dvladir: @bojidaryovchev Yes I prefer `T[]` over `Array<T>`, because it is shorter, on the other hand `ReadonlyArray<T>` is more readable than `readonly T[]` in my opinion. Especially if we could have fields like `readonly someItems: readonly Item[] = []`, such notation looks too complicated in comparison with `readonly someItems: ReadonlyArray<Item> = []`.

projects/step-core/src/lib/modules/basics/shared/repository-parameters.token.ts
```diff
@@ -0,0 +1,7 @@
+import { inject, InjectionToken } from '@angular/core';
+import { RepositoryParametersStorageService } from '../services/repository-parameters-storage.service';
+
+export const REPOSITORY_PARAMETERS = new InjectionToken<ReadonlyArray<string>>('Repository parameters', {
+  providedIn: 'root',
+  factory: () => inject(RepositoryParametersStorageService).parameters,
```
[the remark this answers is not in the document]
dvladir: The idea was to make `RepositoryParametersStorageService` unvaiable to use outside of `@exense/step-core`. It's not exported.
We just know that we have a constant array of parameters, that we can inject and use

projects/step-core/src/lib/modules/table/shared/basic-filter-condition.ts
dvladir: Fixed

projects/step-core/src/lib/modules/table/shared/parameters-filter-condition.ts
dvladir: Fixed

projects/step-core/src/lib/modules/basics/services/repository-parameters-storage.service.ts
bojidaryovchev: well not really, you don't have to remove the ```string[]``` part

bojidaryovchev (approved):

projects/step-core/src/lib/modules/basics/services/repository-parameters-storage.service.ts
dvladir: Can't agree
`let test: string[] = [] as const;` is invalid record
https://www.typescriptlang.org/play?#code/DYUwLgBGIM5gXBOAnAlgOwOYG0C6EBeCPCAQxggGMB7dOAbgCho4A6ABwFcYALACgDkAM2rUBASiY061UK2DVMfFmElA
Typescript shows an error `The type 'readonly []' is 'readonly' and cannot be assigned to the mutable type 'string[]'.`

neogucky (dismissed):

bojidaryovchev (approved):

neogucky (approved):

PR #411, opened 2023-04-05
SED-1970 Scheduler additional screen input filtering does not work

dvladir:


neogucky (approved):
Great generic solution!

bojidaryovchev (approved):

PR #416, opened 2023-04-14
SED-1544 migrate code editor

dvladir:


neogucky (changes requested):
Looks good, but as far as I know there is no place that supports java editor in step. The keyword type script - java only supports uploaded JARs that are too complex to edit them just with the editor.

projects/step-frontend/src/lib/plugins-initializer/register-microfrontend-plugins.ts
```diff
@@ -48,50 +42,21 @@ const loadModule = async (definition: MicrofrontendPluginDefinition): Promise<Pl
...
-    .filter((x) => !!x);
-
-  if (hybridModules.length > 0) {
-    const hostModule = getAngularJSGlobal().module(AJS_MODULE);
-    hostModule.requires.push(...hybridModules);
-  }
+  const compileModules = pluginCtxs.map((ctx) =>
+    compileModule(ctx.definition.entryPoint, ctx.declaration.PluginModule, compileCtx)
+  );
+  const modules = (await Promise.all(compileModules)).filter((x) => !!x) as CompiledModule[];
```
bojidaryovchev, line 59: https://trello.com/c/p2rvxtAO/171-vs-boolean

bojidaryovchev: I'd say for now either leave it as is or perhaps if you like Boolean() more you could use it instead, however I think this is something worth discussing and following everywhere

projects/step-frontend/src/lib/plugins-initializer/shared/module-utils.ts
```diff
@@ -0,0 +1,54 @@
...
+export const registerCompiledModules = async (modules: CompiledModule[]): Promise<void> => {
+  if (!modules.length) {
+    return;
+  }
+
+  const modulesWithInit = modules
+    .map((m) => m.moduleInstance as Partial<PluginOnInit>)
+    .filter((m) => !!m.pluginOnInit) as PluginOnInit[];
+
+  if (modulesWithInit.length > 0) {
```
bojidaryovchev, line 42: > 0

neogucky (dismissed):

bojidaryovchev (approved):

neogucky (approved):

PR #417, opened 2023-04-14
SED-1825 Progress spinner is not displayed when adding / updating / removing keyword input

bojidaryovchev:
- introduced an http interceptor in the angualrjs part
- introduced http-interceptor-bridge.service to expose an observable of intercepted angularjs events
- introduced global-progress-spinner.service to display mat-spinner in a global overlay
- introduced loading.interceptor where we currently handle showing global mat-spinner


dvladir:
@neogucky @bojidaryovchev Please, check my comment in JIRA's ticket

neogucky (approved):
At some point we need to merge our two mechanisms for showing a spinner, but for now this solution works

PR #418, opened 2023-04-17
SED-1547 Mobile support (minimal viable solution)

dvladir:


dvladir:
@neogucky 
> Hm, is there a reason that we did not just use
> 
> `@media only screen and (max-width: 800px) `
> 
> instead of detecting touch devices on code side? Couldn't there be issues for example with PCs that have touch screens connected?

- `css only` solution by checking media query won't work for us, because we need to receive the initial state of the sidebar. Otherwise it will require to additionally check the screen size in the code, that I really prefer to avoid it
- Checking the width doesn't guarantee that mobile device is used
- About desktop touch screen - yes it probably be affected. I think in addition with touch events we can try to check user agent or try to use an additional library that will do the check.

Just to summarize - In my opinion - determine the mobile browser once at application's start is better and optimal approach, in comparison with relying on screen size only.

projects/step-core/src/lib/modules/basics/shared/is-touch-device.token.ts
```diff
@@ -0,0 +1,13 @@
+import { inject, InjectionToken } from '@angular/core';
+import { DOCUMENT } from '@angular/common';
+
+export const IS_TOUCH_DEVICE = new InjectionToken<boolean>('Returns true in case of touch device', {
+  providedIn: 'root',
+  factory: () => {
+    const document = inject(DOCUMENT);
+    return (
+      'ontouchstart' in (document?.defaultView || {}) ||
+      !!(document?.defaultView?.navigator as any)?.['msMaxTouchPoints']
```
bojidaryovchev, line 10:
```
declare global {
  interface Navigator {
    msMaxTouchPoints?: number;
  }
}

export const IS_TOUCH_DEVICE = new InjectionToken<boolean>('Returns true in case of touch device', {
  providedIn: 'root',
  factory: () => {
    const document = inject(DOCUMENT);
    return (
      'ontouchstart' in (document?.defaultView || {}) ||
      !!document?.defaultView?.navigator?.['msMaxTouchPoints']
    );
  },
});
```

dvladir: Not relevant as `IS_TOUCH_DEVICE` has been removed

neogucky (approved):

projects/step-core/src/lib/modules/basics/shared/is-small-screen.token.ts
```diff
@@ -0,0 +1,16 @@
+import { inject, InjectionToken } from '@angular/core';
+import { SCREEN_WIDTH } from './screen-width.token';
+import { distinctUntilChanged, map, Observable } from 'rxjs';
+
+const SMALL_SCREEN_WIDTH = 850;
```
bojidaryovchev, line 5: hmm shouldn't we use the values used by bootstrap / tailwind

image

I think it would be better to use values that are kind of community agreed upon, unless we have a specific case to introduce ones of our own

bojidaryovchev (approved):

projects/step-core/src/lib/modules/basics/shared/is-small-screen.token.ts
dvladir: @bojidaryovchev In comparison here is the bootstrap endpoints
изображение
As you see, it uses absolutely different size breakpoints in comparison with tailwind.
Both of these sets of breakpoints are common in terms of usage of  the concrete style library.
But as we don't use any of these libraries, I think we can define our own dimensions, which are suitable for our needs. 
It's a good idea to standardize it in terms STEP design guide - but I think it should be done under another issue that will be related to global mobile version (not minimal viable solution)

neogucky (approved):

projects/step-core/src/lib/modules/basics/shared/is-small-screen.token.ts
neogucky: Yes for this ticket it is sufficient like this. I agree that we should think about breakpoints (and maybe use those defined by other frameworks) when we do the real mobile version of step.

PR #419, opened 2023-04-18
SED-1975 SED-1924 Fixed error handling

neogucky:


projects/step-frontend/src/lib/modules/_common/interceptors/http-error.interceptor.ts
```diff
@@ -11,18 +18,41 @@ export class HttpErrorInterceptor implements HttpInterceptor {
...
     if (error.status !== 401) {
-      if (error.error?.errorName && error.error?.errorMessage) {
-        this._snackBar.open(error.error.errorName + ': ' + error.error.errorMessage, 'dismiss');
+      if (error.error) {
+        this.showError(error.error);
       } else if (error.name && error.message) {
         this._snackBar.open(error.name + ': ' + error.message, 'dismiss');
       } else {
-        this._snackBar.open(error.error, 'dismiss');
+        this._snackBar.open('Unknown HTTP error', 'dismiss');
```
dvladir, line 26: It looks like these `snakBar` invocations, can be also handled by `showError` if we will add the following condition
```
private showError(error: any): void {
  if (typeof error === 'string') {
     this._snackBar.open(error, 'dismiss')
     return;
  }
  .....
}
```
What do you think?

projects/step-frontend/src/lib/angularjs/js/app.js
```diff
@@ -880,20 +880,28 @@ angular
...
       } else {
         // Legacy error handling
-        if (response.status == 500) {
-          if (
-            responsePayload &&
-            responsePayload.metaMessage &&
-            responsePayload.metaMessage.indexOf('org.rtm.stream.UnknownStreamException') >= 0
-          ) {
+        if (response.status === 500) {
+          if (responsePayload?.metaMessage?.indexOf('org.rtm.stream.UnknownStreamException') >= 0) {
```
bojidaryovchev, line 904: this could be simplified to ```responsePayload?.metaMessage?.includes('org.rtm.stream.UnknownStreamException')```

bojidaryovchev (dismissed):

projects/step-frontend/src/lib/modules/_common/interceptors/http-error.interceptor.ts
neogucky: True, that makes it more elegant

projects/step-frontend/src/lib/angularjs/js/app.js
neogucky: Since it is legacy code anyway I would just leave it. Hopefully it will be removed soon anyway

PR #425, opened 2023-04-20
SED-1975 Improved error handling

neogucky:


dvladir (approved):

projects/step-frontend/src/lib/modules/_common/interceptors/http-error.interceptor.ts
```diff
@@ -11,18 +18,41 @@ export class HttpErrorInterceptor implements HttpInterceptor {
...
+    } else {
+      this._snackBar.open(error, 'dismiss');
+    }
+  }
+
   intercept(req: HttpRequest<any>, next: HttpHandler): Observable<HttpEvent<any>> {
-    return next.handle(req).pipe(catchError((error) => this.handleHttpError(error)));
+    return next.handle(req).pipe(
+      tap((response: HttpEvent<any>) => {
+        return this.handleAsyncError(response);
```
bojidaryovchev, line 53: the ```tap``` operator is used for side effects, a value should not be returned from it; I am not sure if it even works, but it would be better to use do something like ```switchMap((response: HttpEvent<any>) => of(this.handleAsyncError(response))```

neogucky: Good point, I guess tap is ignoring results all together, so we can just remove the return.

neogucky, line 53:
```suggestion
       this.handleAsyncError(response);
```

bojidaryovchev (approved):

PR #438, opened 2023-05-02
SED-1681 support composite in keyword package

dvladir:


projects/step-frontend/src/lib/modules/plan/components/pure-plan-editor/pure-plan-editor.component.html
```diff
@@ -0,0 +1 @@
+<step-plan-editor [id]="id"> </step-plan-editor>
```
neogucky, line 1: I dont like having the "pure plan editor" component so much. I would actually favour to directly use the plan-editor wherever we use the pure-plan-editor. Can you tell me why yout liked it better like this?

neogucky: I like your new naming sugestion, your general approach actually matches how we handle other compoennts so it makes sense to go with it.

projects/step-frontend/src/lib/modules/plan/injectables/pure-plan-edit-api.service.service.ts
```diff
@@ -0,0 +1,64 @@
...
+  loadPlan(id: string): Observable<Plan> {
+    return this._planApi.getPlanById(id);
+  }
+
+  restorePlanVersion(id: string, versionId: string): Observable<Plan> {
+    return this._planApi.restorePlanVersion(id, versionId);
+  }
+
+  savePlan(plan: Plan): Observable<{ id: string; plan: Plan }> {
+    return this._planApi.savePlan(plan).pipe(map((response) => ({ id: response.id!, plan: response })));
```
bojidaryovchev, line 58: given that the response is Plan, why is it explicitly wrapped in this object? I mean it already has an id

projects/step-frontend/src/lib/modules/plan/injectables/pure-plan-edit-api.service.service.ts
```diff
@@ -0,0 +1,64 @@
...
+  private _$location = inject(AJS_LOCATION);
+  private _planApi = inject(PlansService);
+  private _interactiveApi = inject(AugmentedInteractivePlanExecutionService);
+  private _exportDialogs = inject(ExportDialogsService);
+
+  clonePlan(id: string): Observable<Plan> {
+    return this._planApi.clonePlan(id);
+  }
+
+  createRepositoryObjectReference(id: string): RepositoryObjectReference | undefined {
```
bojidaryovchev, line 26: maybe worth declaring it as ```id?: string```

projects/step-frontend/src/lib/modules/plan-editor/injectables/plan-editor-api.service.ts
```diff
@@ -0,0 +1,16 @@
...
+export abstract class PlanEditorApiService {
+  abstract loadPlan(id: string): Observable<Plan>;
+  abstract savePlan(plan: Plan): Observable<{ id: string; plan: Plan }>;
+  abstract clonePlan(id: string): Observable<Plan>;
+  abstract exportPlan(id: string, fileName: string): Observable<void>;
+
+  abstract getPlanHistory(id: string): Observable<History[]>;
+  abstract restorePlanVersion(id: string, versionId: string): Observable<Plan>;
+
+  abstract createRepositoryObjectReference(id: string): RepositoryObjectReference | undefined;
```
bojidaryovchev, line 13: ```id?: string```

projects/step-frontend/src/lib/modules/plan-editor/components/plan-editor/plan-editor.component.ts
```diff
@@ -294,20 +270,20 @@ export class PlanEditorComponent
...
+    this._interactiveSession.execute(this.id!, artefactIds).subscribe(() => {
       if (this.keywords) {
         this.keywords.leafReportsDataSource.reload();
       }
     });
   }
 
-  private loadPlan(planId: string, preselectArtefact?: boolean): void {
-    if (!planId) {
+  private loadPlan(id: string, preselectArtefact?: boolean): void {
```
bojidaryovchev, line 280: ```id?: string```

bojidaryovchev (dismissed):

projects/step-frontend/src/lib/modules/plan/injectables/pure-plan-edit-api.service.service.ts
dvladir: @bojidaryovchev 
If you look at `PurePlanEditApiService` only - it might look strange, but it just implements the common interface, where entity id and plan object are separated.
There is the case of composite keyword, where the identifier is taken from the keyword itself not from the plan.
Check the implementation of `CompositeKeywordPlanApiService`.
That is why id and plan object are need to be split.

neogucky (approved):

bojidaryovchev (dismissed):

neogucky (approved):

PR #439, opened 2023-05-02
SED-1986 When scrolling tables global actions should stay fixed

bojidaryovchev:


projects/step-core/src/lib/modules/table/components/table/table.component.scss
```diff
@@ -1,10 +1,30 @@
-:host {
-  position: relative;
-  display: block;
+step-table {
+  height: 100%;
+  display: flex;
+  flex-direction: column;
```
dvladir, line 5: According to the tasks description - the issue was about controls which are displayed above the table.
On the first glance, the requested behavior can be achieved just by adding a single line in the `host` rule
```
:host {
   position: relative;
   display: block;
  
   // this line is need to be added
   min-width: fit-content;
}
```

What benefits PR's approach gives in comparison with solution, that is shown above?

projects/step-core/src/lib/modules/table/components/table/table.component.scss
```diff
@@ -1,10 +1,30 @@
...
+    flex: 1;
+    overflow: auto;
+    display: block;
+
+    table {
+      width: 100%;
+      border-collapse: separate;
+
+      thead tr {
+        position: sticky;
```
dvladir, line 16: There is no information in the ticket's description, that sticky headers should be implemented. Why it was added?

bojidaryovchev: better UX

neogucky: I actually like it. We just have to be careful in places where there could be multiple tables, but as far as I saw all works well!

neogucky (approved):

PR #442, opened 2023-05-04
SED-1729 Added capability for different labels in dynamic-field-editor

neogucky:


projects/step-core/src/lib/modules/dynamic-forms/components/dynamic-field-editor/dynamic-field-editor.component.ts
```diff
@@ -15,14 +15,18 @@ export class DynamicFieldEditorComponent implements OnChanges {
   @Input() schema?: DynamicFieldsSchema;
   @Input() value?: string;
 
+  @Input() primaryFieldsLabel?: string;
+  @Input() optionalFieldsLabel?: string;
+  @Input() addFieldBtnLabel?: string;
+  @Input() jsonFieldsLabel?: string;
```
dvladir, line 21: As alternative `Inputs (required)`, `Inputs (optional)`, `Add input` and `Input (raw JSON)` could be used as default values for these inputs, instead of `??` in the template.

But as component is downgraded and there is always an assignment from angularJs directive parameters, chosen approach is fine.

dvladir (approved):

dvladir (approved):

PR #443, opened 2023-05-05
SED-1220-Make export compatible with OQL mode

lucianbaciu:


neogucky (dismissed):

projects/step-frontend/src/lib/modules/timeseries/performance-view/filter-bar/filter-bar.component.scss
```diff
@@ -83,12 +79,15 @@ step-ts-filter-bar {
...
+  /**
+  These got override by the theme scss file
+   */
   .apply-btn {
-    color: #4b97de;
-    font-size: 12px;
-    line-height: 28px;
-    margin-right: 12px;
-    font-weight: bold;
+    color: #4b97de !important;
```
bojidaryovchev, line 86: cant u simply rename the class? I cant see it being used elsewhere else

projects/step-frontend/src/lib/modules/timeseries/performance-view/filter-bar/filter-bar.component.ts
```diff
@@ -246,4 +261,33 @@ export class FilterBarComponent implements OnInit, OnDestroy {
...
+    );
+    let selectedTimeRange = this.context.getSelectedTimeRange();
+    let oql = new OQLBuilder()
+      .open('and')
+      .append(`(begin < ${Math.trunc(selectedTimeRange.to)} and begin > ${Math.trunc(selectedTimeRange.from)})`)
+      .append(filtersOql)
+      .append(contextualOql)
+      .build();
+    this.exportInProgress = true;
+    this.timeSeriesService.getMeasurementsAttributes(oql).subscribe((fields) => {
```
bojidaryovchev, line 287: you can say

```
this.timeSeriesService.getMeasurementsAttributes(oql).pipe(switchMap((fields) =>
  this.tableApiService.exportAsCSV('measurements', fields, { filters: [{ oql: oql }] }),
  tap(() => {
    this.exportInProgress = false;
  })
)
.subscribe()
```

projects/step-frontend/src/lib/modules/timeseries/util/oql-builder.ts
```diff
@@ -36,6 +36,6 @@ export class OQLBuilder {
   }
 
   build(): string {
-    return this.clauses.join(this.separator);
+    return this.clauses.join(' ' + this.separator + ' ');
```
bojidaryovchev, line 39: separator is set in the open method, I think this change will break something

bojidaryovchev (approved):

projects/step-frontend/src/lib/modules/timeseries/util/oql-builder.ts
lucianbaciu: it is also initialize with 'and'

projects/step-frontend/src/lib/modules/timeseries/performance-view/filter-bar/filter-bar.component.scss
neogucky: Could you in stead use a more concrete selector like "div>button.apply-btn"?

Also, we should use a variable for the colour.

neogucky (approved):
Approved, but please still look at the issue with the selector / colour

projects/step-frontend/src/lib/modules/timeseries/performance-view/filter-bar/filter-bar.component.scss
lucianbaciu: Yes, will do

neogucky (approved):

projects/step-frontend/src/lib/modules/timeseries/performance-view/filter-bar/filter-bar.component.ts
```diff
@@ -246,4 +261,35 @@ export class FilterBarComponent implements OnInit, OnDestroy {
     }
     return false;
   }
+
+  exportRawData() {
+    if (this.exportInProgress) {
+      return;
+    }
```
bojidaryovchev, line 268: it's a good practice in terms of clean code to leave an empty line before/after if/for/etc.

projects/step-frontend/src/lib/modules/timeseries/util/filter-utils.ts
```diff
@@ -7,47 +7,56 @@ export class FilterUtils {
     );
   }
 
-  static objectToOQL(object: { [key: string]: string }, attributesPrefix?: string) {
+  static objectToOQL(
+    object: { [key: string]: string },
```
bojidaryovchev, line 11: another way to define this would be ```object: Record<string, string>``` - no need to modify it tho

projects/step-frontend/src/lib/modules/timeseries/util/filter-utils.ts
```diff
@@ -7,47 +7,56 @@ export class FilterUtils {
...
+    attributesPrefix?: string,
+    attributeProcessFn?: (x: string) => string
```
bojidaryovchev, line 13: try to avoid having names like ```x```, instead try to use the actual name of the thing (looks like the name should be ```attribute```)

projects/step-frontend/src/lib/modules/timeseries/util/oql-builder.ts
bojidaryovchev: this is not correct tho, I can see that the ```OQLBuilder``` is used only here

image

however, keep in mind we should be able to unit test each piece individually, so consider the following scenario

```const foo = new OQLBuilder().open('-').append('a').append('b').build();```

now, when I try to test it

```expect(foo).toBe('a-b');```

the test will fail, because the actual result will be ```'a - b'```

bojidaryovchev: anyway, up to u to decide, I would personally leave the user provide the proper separator from the outside

neogucky (approved):

PR #445, opened 2023-05-05
SED-1990-Default interval can be passed in the url, and other fixes f…

lucianbaciu:
…or previous tasks


projects/step-core/src/lib/services/dashboard.service.ts
```diff
@@ -5,6 +5,8 @@ import { ViewRegistryService } from './view-registry.service';
 import { downgradeInjectable, getAngularJSGlobal } from '@angular/upgrade/static';
 import { AJS_MODULE } from '../shared';
 
+const ONE_DAY_MS = 86400000;
```
neogucky, line 8:
```suggestion
const ONE_DAY_MS = 24 * 60 * 60 * 1000;
```
I would write it like this.

projects/step-frontend/src/lib/modules/timeseries/analytics/analytics-page.component.ts
```diff
@@ -50,20 +48,27 @@ export class AnalyticsPageComponent implements OnInit, OnDestroy {
...
+        const range = Number(urlParams.relativeRange);
+        this.timeRangeSelection = this.getClosestRangeOption(range);
+      } else {
+        this.timeRangeSelection = this.timeRangeOptions[0];
+      }
+      start = now - this.timeRangeSelection.relativeSelection!.timeInMs;
       end = now;
-      this.timeRangeSelection = this.timeRangeOptions[0];
     }
+    console.log(end - start);
```
neogucky, line 66:
```suggestion
```

projects/step-frontend/src/lib/modules/timeseries/performance-view/filter-bar/filter-bar.component.ts
```diff
@@ -184,18 +176,18 @@ export class FilterBarComponent implements OnInit, OnDestroy {
...
-    const filterIndex = this.visibleFilters.findIndex((i) => i.attributeName === item.attributeName);
+    const filterIndex = this.filterItems.findIndex((i) => i.attributeName === item.attributeName);
 
     if (filterIndex !== -1) {
       const index =
-        filterIndex < this.visibleFilters.length - this.visibleFilters.length
+        filterIndex < this.filterItems.length - this.filterItems.length
           ? filterIndex
-          : filterIndex + this.visibleFilters.length;
+          : filterIndex + this.filterItems.length;
```
neogucky, line 190: Always put calculations in parantheses to make it clear in which order they are compared / calculated

projects/step-frontend/src/lib/modules/timeseries/time-series-context-params.ts
```diff
@@ -5,6 +5,9 @@ export interface TimeSeriesContextParams {
   id: string;
   timeRange: TSTimeRange;
   grouping: string[];
-  baseFilters: any;
+  /**
+   * @deprecated
+   */
+  baseFilters?: any;
```
neogucky, line 11: Would removing the deprecated places create to many changes to be integrated into this PR? If yes, I would create a ticket to fix it and reference the ticket ID here

projects/step-frontend/src/lib/modules/timeseries/util/filter-utils.ts
```diff
@@ -26,7 +26,10 @@ export class FilterUtils {
     }
     let andFilters: (string | undefined)[] = items.map((item) => {
       let clause;
-      const attributeName = attributesPrefix ? `${attributesPrefix}.${item.attributeName}` : item.attributeName;
+      let attributeName = attributesPrefix ? `${attributesPrefix}.${item.attributeName}` : item.attributeName;
+      if (attributeName.includes('')) {
```
neogucky, line 30: Can you quickly tell me what this does? I would assume .includes('') is always true except if it is not a string (and then it will throw an error)

projects/step-frontend/src/lib/modules/timeseries/performance-view/filter-bar/filter-bar.component.ts
bojidaryovchev: I think the parentheses are getting removed by Prettier as they are not technically needed

bojidaryovchev (dismissed):

projects/step-frontend/src/lib/modules/timeseries/time-series-context-params.ts
lucianbaciu: I can check

projects/step-frontend/src/lib/modules/timeseries/util/filter-utils.ts
lucianbaciu: @neogucky it had to be " " (white space). I fixed it

projects/step-frontend/src/lib/modules/timeseries/time-series-context-params.ts
lucianbaciu: @neogucky it can have multiple implications and an overall testing is also needed. I created https://exense.atlassian.net/browse/SED-2042

bojidaryovchev (approved):

PR #446, opened 2023-05-09
SED-2014 message component implementation

neogucky:


projects/step-core/src/lib/modules/table/shared/table-remote-data-source.ts
```diff
@@ -150,19 +150,22 @@ export class TableRemoteDataSource<T> implements TableDataSource<T> {
...
 
-  private typeFilter?: { [key: string]: SearchValue };
+  private filters: { [key: string]: SearchValue } = {};
 
   constructor(
     private _tableId: string,
     private _rest: TableApiWrapperService,
     private _requestColumnsMap: { [key: string]: string },
-    private _typeFilter?: string[]
+    private _filters?: { [key: string]: string[] }
```
bojidaryovchev, line 159: personal preference - ```_filters?: Record<string, string[]>```

projects/step-core/src/lib/modules/table/shared/table-remote-data-source.ts
```diff
@@ -182,8 +185,8 @@ export class TableRemoteDataSource<T> implements TableDataSource<T> {
   }
 
   private createInternalRequestObject({ params, filter, search }: TableFilterOptions = {}): TableRequestInternal {
-    if (this.typeFilter) {
-      search = { ...(search || {}), ...this.typeFilter };
+    if (Object.keys(this.filters).length > 0) {
```
bojidaryovchev, line 188: personal preference - ```if (Object.keys(this.filters).length) {```

bojidaryovchev (dismissed):

bojidaryovchev (approved):

projects/step-core/src/lib/modules/table/shared/table-remote-data-source.ts
neogucky: I will add it for discussion to Trello, I don't have a preference but if we want to go with one of the two we should do it everywhere (i.e. also for "_requestColumnsMap" above this)

PR #455, opened 2023-05-12
SED-2000 [OS] Migrate function configuration dialog

bojidaryovchev:


projects/step-core/src/lib/client/augmented/services/augmented-keywords-service.ts
```diff
@@ -21,7 +22,36 @@ export class AugmentedKeywordsService extends KeywordsService {
...
-  constructor(override httpRequest: BaseHttpRequest, private _tableRest: TableApiWrapperService) {
-    super(httpRequest);
+  override newFunctionTypeConf(type: string): Observable<Function>;
+  override newFunctionTypeConf(type: string, serviceRoot: string): Observable<Function>;
+  override newFunctionTypeConf(type: string, serviceRoot?: string): Observable<Function> {
+    return this.httpRequest.request({
+      method: 'GET',
+      url: `/${serviceRoot}/types/${type}`,
+    });
+  }
```
dvladir, line 32: @neogucky Something wrong with this approach, that was originally implemented.
We shouldn't change that path the root path of the api services.
Previously it was done for `masksList` in pdf-test plugin, but EE has it's own endpoint to invoke these methods.
So in case of plugin another service should be used, instead of changing path in the keyword service

@neogucky, @bojidaryovchev 
Is it possible to rollback the `function-dialogs.service` to direct `httpClient` usage, and then fix this problem correctly under the separate task?

neogucky: This is approach is based on the original approach used in the AJS version. I agree that calling the correct endpoint without a wild card would be preferrable, also we should limit augmenting things whenever we can because this undermines the Swagger concept.

I don't get the point regarding EE, the service is in the core so should work fine with both versions. Can you elaborate @dvladir ?

projects/step-core/src/lib/services/export-dialogs.service.ts
```diff
@@ -1,15 +1,15 @@
...
 
 @Injectable({
   providedIn: 'root',
 })
 export class ExportDialogsService {
-  constructor(private _uibModalHelper: UibModalHelperService) {}
+  private _uibModalHelper = inject(UibModalHelperService);
 
-  displayExportDialog<T>(title: string, path: string, filename: string): Observable<T> {
+  displayExportDialog(title: string, path: string, filename: string): Observable<boolean | void> {
```
dvladir, line 12: Such return type looks strange.
If we're only interested it, will the error be returned or not - than it should be boolean only

projects/step-core/src/lib/shared/function-type-form-component.ts
```diff
@@ -0,0 +1,48 @@
...
+export abstract class FunctionTypeFormComponent<T extends FormGroup> implements CustomComponent, OnInit, OnDestroy {
+  private readonly terminator$ = new Subject<void>();
+
+  protected abstract formGroup: T;
+  protected abstract formGroupValidator: ValidatorFn;
+
+  context!: FunctionTypeFormComponentContext;
+
+  ngOnInit(): void {
+    this.context!.formGroup.addValidators(this.formGroupValidator);
```
dvladir, line 19: Custom subform management looks complicated.
This could be achieved by adding/removing subform dynamically to the parent form group.
In that case validation and validity update will work out of the box.
Probably the setValueToForm / setValueToModel could be also simplified.

Additional ticket will be created to improve this mechanism.

dvladir (changes requested):

projects/step-frontend/src/lib/modules/function/components/function-configuration-dialog/function-configuration-dialog.component.ts
```diff
@@ -0,0 +1,168 @@
+import { ChangeDetectorRef, Component, inject, OnDestroy, OnInit } from '@angular/core';
+import { FormBuilder } from '@angular/forms';
+import { MatDialogRef, MAT_DIALOG_DATA } from '@angular/material/dialog';
+import {
+  AJS_LOCATION,
+  AlertType,
+  AugmentedKeywordsService,
+  AuthService,
+  DialogsService,
+  Function,
```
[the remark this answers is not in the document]
bojidaryovchev: I've taken the effort to use the alias everywhere

projects/step-core/src/lib/shared/function-type-composite.form.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export const functionTypeCompositeFormSetValueToForm = (
+  form: FunctionTypeCompositeForm,
+  model: FunctionComposite
+): void => {};
+
+export const functionTypeCompositeFormSetValueToModel = (
+  form: FunctionTypeCompositeForm,
+  model: FunctionComposite
+): void => {};
```
dvladir, line 20: @bojidaryovchev 
What is the purpose of these functions above?
They actually do nothing

projects/step-core/src/lib/shared/function-configuration-dialog.form.ts
```diff
@@ -0,0 +1,98 @@
...
+    };
+  }
+
+  return null;
+};
+
+const DEFAULT_SCHEMA_VALUE = '{}';
+const DEFAULT_CALL_TIMEOUT_MS = 180000;
+
+export const functionConfigurationDialogFormCreate = (
```
dvladir, line 39: Most of these forms are belong to the concrete components and don's used anywhere else.
From this point of view, they can be considered as the part of the concrete components. I think it make sense to move the to corresponding component's folders.

dvladir (dismissed):

neogucky (approved):

neogucky (dismissed):

neogucky (dismissed):

PR #457, opened 2023-05-17
SED-2053 add explicit execution result column in executions list view

neogucky:


projects/step-frontend/src/lib/modules/execution/components/execution-result/execution-result.component.html
```diff
@@ -1,4 +1,3 @@
-<span [matTooltip]="displayResult" [class]="statusClass">
-  <step-icon [name]="icon" [class.spin-animation]="icon === 'refresh-cw'">
-  </step-icon>
+<span [matTooltip]="displayResult" [ngClass]="[statusClass, 'step-icon-' + result]">
```
dvladir, line 1: What is the reason of this change, if `statusClass` itself will contain 'step-icon-' + result.
Check line 33 in component's code

neogucky: Actually it contained ".step-icon-" + result that's why it did not work and I did not think it might actually contain what I needed. Fixed it now!

bojidaryovchev (approved):

PR #458, opened 2023-05-19
SED-1997 [OS] migrate new scheduler task dialog

dvladir:


projects/step-core/src/lib/modules/basics/components/cron-selector/cron-selector.component.ts
```diff
@@ -1,22 +1,62 @@
...
-export class CronSelectorComponent {
+export class CronSelectorComponent implements ControlValueAccessor, OnDestroy {
+  readonly _cronPresets = inject(CRON_PRESETS);
+  readonly trackByKeyValue: TrackByFunction<KeyValue<string, string>> = (index, item) => item.key;
+
   @Input() label?: string;
-  @Input() cronString?: string;
+  @Input() placeholder?: string;
+
+  private onChange: OnChange = noop;
```
bojidaryovchev, line 22: we could completely avoid using ```noop``` if we declared it like ```private onChange?: OnChange;``` and then invoked it like ```this.onChange?.('foo')```, also not quite sure what's the purpose of setting it back to ```noop``` in ```ngOnDestroy()```

bojidaryovchev (approved):

projects/step-core/src/lib/modules/basics/components/cron-selector/cron-selector.component.ts
bojidaryovchev: I mean it looks strange to me because they are after all component properties so it would make sense when the component gets destroyed and therefore not referenced for its properties to be garbage-collected

neogucky (approved):

PR #459, opened 2023-05-22
SED-1998 [OS] Migrate edit scheduler task dialog

dvladir:


projects/step-core/src/lib/components/select-plan/select-plan.component.html
```diff
@@ -1,7 +1,7 @@
-<step-form-field>
-  <step-label show-required-marker>Plan</step-label>
-  <input type="text" [formControl]="planNameControl" [tabIndex]="-1" />
+<step-form-field [class.ng-invalid]="_ngControl.invalid" [class.ng-touched]="_ngControl.touched">
+  <step-label [attr.show-required-marker]="showRequiredMarker || null">Plan</step-label>
+  <input type="text" readonly [disabled]="isDisabled" [ngModel]="planValue?.attributes?.['name']" [tabIndex]="-1" />
   <step-suffix class="button-suffix">
-    <button type="button" mat-stroked-button (click)="selectPlan()" [disabled]="disabled">Select</button>
+    <button type="button" mat-stroked-button (click)="selectPlan()" [disabled]="isDisabled">Select</button>
```
bojidaryovchev, line 5: also for the coding style guide - ```disabled``` vs ```isDisabled```

projects/step-core/src/lib/components/edit-scheduler-task-dialog/edit-scheduler-task-dialog.component.html
```diff
@@ -0,0 +1,61 @@
...
+      </step-form-field>
+    </form>
+    <mat-expansion-panel [(expanded)]="showParameters">
+      <mat-expansion-panel-header>
+        <mat-panel-title>Advanced</mat-panel-title>
+      </mat-expansion-panel-header>
+      <section class="advanced-body">
+        <step-form-field>
+          <step-label>Parameters</step-label>
+          <textarea
```
bojidaryovchev, line 46: I think this should use a reactive form and have a JSON validator which would then display an error below if the JSON is invalid (I've done that as part of the function configuration dialog migration)

dvladir: @bojidaryovchev This editor uses `step-custom-froms` component, that is downgraded, is used in legacy UI and doesn't support reactive forms. That's why I decided to use template driven form for this dialog.

projects/step-core/src/lib/components/edit-scheduler-task-dialog/edit-scheduler-task-dialog.component.html
```diff
@@ -0,0 +1,61 @@
+<step-modal-window [title]="isNew ? 'New task' : 'Edit task'">
+  <div step-modal-window-body>
+    <step-alert *ngIf="error" type="danger">
```
bojidaryovchev, line 3: another thing we should consider - say we have an ```enum Foo { BAR = 'bar' }``` and component ```app-something``` which has an ```@Input() foo!: Foo;```

so should we do it like ```<app-something foo="bar"></app-something``` or ```<app-something [foo]="Foo.BAR"></app-something>``` (having ```protected readonly Foo = Foo``` in the ```.ts```)

projects/step-core/src/lib/components/edit-scheduler-task-dialog/edit-scheduler-task-dialog.component.html
bojidaryovchev: @dvladir I also used this component in the function configuration dialog

bojidaryovchev (approved):

projects/step-core/src/lib/components/edit-scheduler-task-dialog/edit-scheduler-task-dialog.component.html
bojidaryovchev: image

bojidaryovchev: image

bojidaryovchev: image

bojidaryovchev: image

bojidaryovchev (approved):

PR #467, opened 2023-05-24
SED-1310 Find solution to run scripts on all host systems safely

dvladir:


neogucky (approved):
Looks good, works on OSX

PR #470, opened 2023-05-25
SED-1994 [OS] migrate the edit resource Dialog

bojidaryovchev:


dvladir (changes requested):

projects/step-core/src/lib/components/predefined-options-input/predefined-options-input.component.ts
```diff
@@ -0,0 +1,63 @@
...
+  styleUrls: ['./predefined-options-input.component.scss'],
+})
+export class PredefinedOptionsInputComponent implements ControlValueAccessor {
+  @Input() label?: string;
+  @Input() placeholder?: string = '';
+  @Input() predefinedOptions!: KeyValue<string, string>[];
+  @Input() predefinedOptionsLabel!: string;
+  @Input() showRequiredMarker?: boolean;
+  @Input() isParentInvalid?: boolean;
+  @Input() isParentTouched?: boolean;
```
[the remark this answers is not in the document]
bojidaryovchev: the idea is that if we have a component that implements ```ControlValueAccessor``` we should be able to use it within other such components, these 2 properties make it possible for it to be controlled from the outside when necessary; imo, we should account for that in all ```ControlValueAccessor``` components so that they are composable

projects/step-frontend/src/lib/modules/resources/components/resource-configuration-dialog/resource-configuration-dialog.form.ts
```diff
@@ -0,0 +1,50 @@
...
+  const { attributes, resourceType } = model;
+
+  if (attributes?.['name']) {
+    form.patchValue({
+      name: attributes['name'],
+      content: createDynamicValueString({
+        value: attributes['name'],
+      }),
+    });
+    form.controls.name.disable();
```
dvladir, line 32: @bojidaryovchev Why assigning the value, invokes form's control disable?

projects/step-frontend/src/lib/modules/resources/components/resource-configuration-dialog/resource-configuration-dialog.component.ts
```diff
@@ -0,0 +1,149 @@
...
+    // Preserve the original onFilesChange
+    this.onFilesChangeRef = this.resourceInput!.onFilesChange.bind(this.resourceInput);
+
+    // Override onFilesChange
+    // - given the way the resource input is written,
+    // - it requires isResource to be false for the template
+    // - in order to render the same content as the old version,
+    // - and at the same time isResource needs to be true for
+    // - the onFilesChange handler to perform the proper action
+    this.resourceInput!.onFilesChange = this.onFilesChangeRefOverride.bind(this);
```
dvladir, line 133: In case if the code above is required, because `resource.attributes.name` doesn't contain `resource:`, it can be solved in the following way:
In component's body
```
readonly stModel$ = this.formGroup.controls.content.valueChanges
   .pipe(
      startWith(this.formGroup.controls.content.value),
      map((content) => content?.value),
      map((value) => {
           if (!value) {
                return value;
           }
           if (!value.startsWith('resource:')) {
                return `resource:${value}`;
           }
           return value;
      })
   )

```
In component's template:
```
      <step-resource-input
        #resourceInput
        label="Content"
        [stType]="formGroup.controls.resourceType.value"
        [uploadOnly]="true"
        [disableServerPath]="true"
        [stModel]="stModel$ | async"
        (stModelChange)="onContentChange($event)"
        (filesChange)="onFilesChange()"
        (uploadComplete)="onUploadComplete()"
      ></step-resource-input>

```
In that case `resourceInput` manual initialization and external method's override won't required.

dvladir (changes requested):
Avoid the runtime method overrides of child component's instance in the parent's component.

dvladir (approved):

PR #472, opened 2023-05-25
Sed 2030 finalize UI/ux design step 22

lucianbaciu:


projects/step-frontend/src/lib/modules/timeseries/performance-view/filter-bar/filter-bar.component.html
```diff
@@ -1,8 +1,7 @@
+<div class="warning-mark" *ngIf="rawMeasurementsModeActive">
+  You are using custom filters which may slow down the dashboard.
+</div>
```
dvladir, line 3:
```suggestion
<step-alert type="warning" *ngIf="rawMeasurementsModeActive">
  You are using custom filters which may slow down the dashboard.
</step-alert>
```

projects/step-frontend/src/lib/modules/timeseries/performance-view/filter-bar/filter-bar.component.scss
```diff
@@ -11,22 +11,29 @@ step-ts-filter-bar {
...
+    display: flex;
+    margin-top: 12px !important;
   }
 
   .filters {
     display: flex;
     gap: 12px;
     padding: 12px;
     align-items: center;
+    height: 68px;
```
dvladir, line 36: We've decided to use `rem` instead of `px` for styling non material parts. Is it possible to replace it in this style sheet?

bojidaryovchev (approved):

projects/step-frontend/src/lib/modules/timeseries/chart/time-series-chart.component.ts
```diff
@@ -26,7 +26,7 @@ import uPlot = require('uplot');
   styleUrls: ['./time-series-chart.component.scss'],
 })
 export class TimeSeriesChartComponent implements OnInit, AfterViewInit, OnChanges {
-  private readonly HEADER_WITH_FOOTER_SIZE = 80;
+  private readonly HEADER_WITH_FOOTER_SIZE = 48;
   readonly WRAPPER_PADDING = '12px';
```
neogucky, line 30: This was not newly introduced but having a style hardcoded here seems weird. Why is this not something defined by the styles?

neogucky: Ah it seems that you removed it in the template, so we can remove it here as well now?

lucianbaciu: yes, we can remove it. I will take care of it

PR #480, opened 2023-05-31
SED-2051 display warning when editing composite from keyword package

neogucky:


projects/step-frontend/src/lib/modules/function/components/composite-function-editor/composite-function-editor.component.ts
```diff
@@ -17,6 +17,19 @@ import { CompositeKeywordPlanApiService } from '../../services/composite-keyword
...
+  protected composite?: KeywordFunction;
+
+  ngOnChanges(changes: SimpleChanges): void {
+    const cId = changes['id'];
+    if (cId?.previousValue !== cId?.currentValue || cId?.firstChange) {
+      this._functionApiService.getFunctionById(cId.currentValue).subscribe((keywordFunction) => {
+        this.composite = keywordFunction;
+      });
+    }
+  }
```
dvladir, line 32:
```suggestion
  ngOnChanges(changes: SimpleChanges): void {
    const cId = changes['id'];
    if (cId?.previousValue !== cId?.currentValue || cId?.firstChange) {
     this.loadKeyword(cId?.currentValue);
    }
  }
  
  private loadKeyword(id?: string): void {
      if (!id) {
         this.composite = undefined;
         return;
      }
      this._functionApiService.getFunctionById(id).subscribe((keywordFunction) => {
        this.composite = keywordFunction;
      });
  }
```

dvladir (dismissed):

PR #493, opened 2023-06-11
SED-2134-Ability to check/uncheck all series selection

lucianbaciu:


neogucky (approved):
The implementation is fine, but did you check the step-bulk-operations directive? Potentially it could also be used for this. If you prefer the more basic solution it is also fine.

projects/step-frontend/src/lib/modules/timeseries/performance-view/table/timeseries-table.component.ts
```diff
@@ -84,6 +90,10 @@ export class TimeseriesTableComponent implements OnInit, OnDestroy {
     this.keywordsService.toggleKeyword(keyword);
   }
 
+  onAllSeriesCheckboxClick(event: any) {
```
bojidaryovchev, line 93: we dont seem to care about the event, therefore it can be removed as a parameter

lucianbaciu:
> The implementation is fine, but did you check the step-bulk-operations directive? Potentially it could also be used for this. If you prefer the more basic solution it is also fine.

I will check it anyway. here I think we can leave it basic. we have the same basic approach in another component too.

PR #500, opened 2023-06-14
SED-2176 os migrate plan selection list

neogucky:


projects/step-frontend/src/lib/modules/plan/components/plan-selection/plan-selection.component.scss
```diff
@@ -0,0 +1,4 @@
+.mat-column-singleSelect,
+.mat-column-bulkSelection {
+  width: 48px;
```
bojidaryovchev, line 3: I see this in a 2nd PR, does it make sense to have it as a global style?

bojidaryovchev (approved):

projects/step-frontend/src/lib/modules/plan/components/plan-selection/plan-selection.component.scss
neogucky: Even if it duplicates code I would leave it here since it directly influences how the table is rendered.

PR #518, opened 2023-07-03
SED-2165 Reloading plan on out-of-sync error

neogucky:


projects/step-frontend/src/lib/modules/plan/injectables/pure-plan-edit-api.service.ts
```diff
@@ -54,8 +56,18 @@ export class PurePlanEditApiService implements PlanEditorApiService {
     return this._planApi.restorePlanVersion(id, versionId);
   }
 
-  savePlan(plan: Plan): Observable<{ id: string; plan: Plan }> {
-    return this._planApi.savePlan(plan).pipe(map((response) => ({ id: response.id!, plan: response })));
+  savePlan(plan: Plan): Observable<{ id: string; plan: Plan; forceRefresh?: boolean }> {
+    return this._planApi.savePlan(plan).pipe(
+      map((response) => ({ id: response.id!, plan: response })),
+      catchError((e: any) => {
```
bojidaryovchev, line 62: we dont seem to need to make use of the error, so it can be removed

bojidaryovchev, line 62:
```suggestion
      catchError(() => {
```

bojidaryovchev (approved):

projects/step-frontend/src/lib/modules/plan/injectables/pure-plan-edit-api.service.ts
neogucky: Thanks!

PR #548, opened 2023-08-10
SED-2357 Selected keyword type not displayed when navigating back to keyword list

dvladir:
* SED-2357 Selected keyword type not displayed when navigating back to keyword list

* SED-2357 Fix dropdowns for custom columns

### DIFF
--- a/projects/step-core/src/lib/modules/basics/components/array-filter/array-filter.component.ts
+++ b/projects/step-core/src/lib/modules/basics/components/array-filter/array-filter.component.ts
@@ -66,10 +66,20 @@ export class ArrayFilterComponent<T = unknown> extends BaseFilterComponent<strin
   }
 
   protected override transformFilterValueToControlValue(value: string): unknown {
+    let result: string[];
+
     if (value.startsWith('(') && value.endsWith(')') && value.includes('|')) {
-      return value.substring(1, value.length - 1).split('|');
+      result = value.substring(1, value.length - 1).split('|');
+    } else {
+      result = [value];
     }
-    return [value];
+
+    return result.map((value) => {
+      if (value.startsWith('^') && value.endsWith('$')) {
+        return value.substring(1, value.length - 1);
+      }
+      return value;
+    });
   }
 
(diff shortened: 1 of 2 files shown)


projects/step-core/src/lib/modules/basics/components/array-filter/array-filter.component.ts
```diff
@@ -66,10 +66,20 @@ export class ArrayFilterComponent<T = unknown> extends BaseFilterComponent<strin
   }
 
   protected override transformFilterValueToControlValue(value: string): unknown {
+    let result: string[];
+
     if (value.startsWith('(') && value.endsWith(')') && value.includes('|')) {
```
jeromecomte, line 71: @dvladir In addition to the simple FieldFilter based on regex that is used here, the table service now supports so-called CollectionFilter. CollectionFilters are hierarchical filters that support operators like Or, And, etc. Using this to build up the filter passed as payload to the service rest/table/functions would also improve the handling in this function

{"collectionFilter":{"type":"Or", "children": [{"type":"Equals","field":"type","expectedValue":"step.plugins.functions.types.CompositeFunction"},{"type":"Equals","field":"type","expectedValue":"step.plugins.java.GeneralScriptFunction"}]}}

jeromecomte (approved):

projects/step-core/src/lib/modules/basics/components/array-filter/array-filter.component.ts
dvladir: @jeromecomte We're already using CollectionFilter for some specific cases, were simple string value or regexp can not be applied.
For example: Date filters on the execution page, scope filter on the parameters page, etc.

`Keyword Type filter` was inherited from `Array Filter`, I guess it was the fastest solution.
But in theory we can change it to use the CollectionFilter instead of regex, as you described it above.

jeromecomte:
Great. Wouldn't change this for the next minor (branch 3.22). I let you
decide if it make sense do it or not on master for the future.

PR #556, opened 2023-08-28
SED-2373 Listing all API keys and giving possibility to revoke

neogucky:


projects/step-core/src/lib/modules/basics/shared/generate-api-key-strategy.ts
```diff
@@ -1,3 +1,7 @@
+import { Observable } from 'rxjs';
+
 export interface GenerateApiKeyStrategy {
-  showGenerateApiKeyDialog(): void;
+  showGenerateApiKeyDialog(): Observable<any>;
+  getServiceAccountTokens(): Observable<Array<any>>;
+  revoke(id: string): Observable<any>;
```
dvladir, line 6:
```suggestion
  revoke(id: string): Observable<void>;
```
As I see, we're not interested in the result value of this method and don't analyze it. Let's make it `void`

projects/step-core/src/lib/modules/basics/shared/generate-api-key-strategy.ts
```diff
@@ -1,3 +1,7 @@
+import { Observable } from 'rxjs';
+
 export interface GenerateApiKeyStrategy {
-  showGenerateApiKeyDialog(): void;
+  showGenerateApiKeyDialog(): Observable<any>;
+  getServiceAccountTokens(): Observable<Array<any>>;
```
dvladir, line 5: The template below shows, that we know something about token's structure.
At least it should contain: id, label, creationTime and expirtyTime.
It would be nice to define such interface and use it in OS.
The method notation it this case could be written as generic:
```
getServiceAccountTokens<T extends BaseAccountToken>(): Observable<T[]>;
```

projects/step-frontend/src/lib/modules/admin/components/my-account/my-account.component.html
```diff
@@ -26,6 +26,30 @@ <h4>Credentials</h4>
...
+        <thead>
+          <tr>
+            <th>Label</th>
+            <th>Created</th>
+            <th>Valid until</th>
+            <th>Action</th>
+          </tr>
+        </thead>
+        <tbody>
+          <tr *ngFor="let token of tokens">
```
dvladir, line 39: Add `trackBy`

projects/step-core/src/lib/modules/basics/services/generate-api-key.service.ts
```diff
@@ -11,10 +12,24 @@ export class GenerateApiKeyService implements GenerateApiKeyStrategy {
     this.strategy = strategy;
   }
 
-  showGenerateApiKeyDialog(): void {
+  showGenerateApiKeyDialog(): Observable<any> {
```
dvladir, line 15: Interface doesn't implemented properly. Service's method return types should be the same as in strategy

dvladir (changes requested):

dvladir (approved):

dvladir (dismissed):

PR #562, opened 2023-08-31
SED-2345 Use correct id for composite version history

neogucky:


dvladir (approved):

projects/step-frontend/src/lib/modules/plan-editor/components/plan-editor-base/plan-editor-base.component.ts
```diff
@@ -154,19 +167,42 @@ export class PlanEditorBaseComponent
...
+        mergeMap((currentVersion) =>
+          this._restoreDialogsService.showRestoreDialog(currentVersion, versionHistory, permission).pipe(
+            filter((restoreVersion) => restoreVersion !== undefined),
+            mergeMap((restoreVersion) => this._planEditorApi.restorePlanVersion(this.id || plan.id!, restoreVersion)),
+            catchError((error) => {
+              console.error(error);
+              return EMPTY;
+            })
+          )
+        )
```
dvladir, line 205: Looks like that inner `pipe` is redundant.
Also I prefer to use the `switchMap` instead of `mergeMap`, because `switchMap` guarantees that previous observable has been completed. Here it's not significant because we have several sequenced requests, but in some cases `mergeMap` might be the source of memory leaks (ex. If the request is performed on each user input. With mergeMap each new request won't complete the previous one. Plus If the result observable will be shared for several subscriptions, there might be issues)

```suggestion
        switchMap((currentVersion) => this._restoreDialogsService.showRestoreDialog(currentVersion, versionHistory, permission)),
        filter((restoreVersion) => restoreVersion !== undefined),
        switchMap((restoreVersion) => this._planEditorApi.restorePlanVersion(this.id || plan.id!, restoreVersion)),
        catchError((error) => {
            console.error(error);
            return EMPTY;
        })
```

dvladir (changes requested):

dvladir (approved):

PR #563, opened 2023-08-31
SED-1838 edit function package original resource removed when cancelling update

neogucky:


projects/step-core/src/lib/modules/resource-input/components/resouce-input/resouce-input.component.ts
```diff
@@ -294,20 +295,32 @@ export class ResourceInputComponent implements OnInit, OnChanges, OnDestroy {
...
     }
 
-    this._augmentedResourcesService.deleteResource(this.lastUploadedResourceId).subscribe(() => {
-      delete this.lastUploadedResourceId;
+    const requests: Observable<any>[] = [];
+    const cleanupIds = Array.from(this.uploadedResourceIds);
+
+    for (const id of cleanupIds) {
+      requests.push(this._augmentedResourcesService.deleteResource(id));
+    }
```
dvladir, line 318: Not critical, but personally I would solve it in this way
```suggestion
    const cleanupIds = [...this.uploadedResourceIds];
    const requests = cleanupIds.map((id) => this._augmentedResourcesService.deleteResource(id));
```

dvladir (approved):

PR #573, opened 2023-09-07
SED-2343 execution picker for analytics filter

lucianbaciu:


lucianbaciu:
This is just a draft (will be done today (Thrusday) most likely).

To be done:
- rem instead of px
- @inject instead of constructor
- cleanup

projects/step-core/src/lib/modules/entity/components/base-entity-selection-table/base-entity-selection-table.component.ts
```diff
@@ -33,11 +33,11 @@ export abstract class BaseEntitySelectionTableComponent implements AfterViewInit
     this.cleanupContext();
   }
 
-  handleSelect(id: string): void {
+  handleSelect(item: any): void {
```
neogucky, line 36: we should avoid any, what is the expected type here?

projects/step-core/src/lib/modules/entity/components/select-entity-of-type/select-entity-of-type.component.ts
```diff
@@ -20,7 +20,7 @@ export class SelectEntityOfTypeComponent implements OnInit, OnDestroy {
   protected entity = this._entityRegistry.getEntityByName(this._dialogData.entityName);
   protected selectEntityContext: SelectEntityContext = {
     multipleSelection: !this._dialogData.singleSelection,
-    handleSelect: (selectedId: string) => this.handleSelect(selectedId),
+    handleSelect: (item: any) => this.handleSelect(item),
```
neogucky, line 23: again "any"

projects/step-core/src/lib/modules/entity/components/select-entity-of-type/select-entity-of-type.component.ts
```diff
@@ -72,10 +72,10 @@ export class SelectEntityOfTypeComponent implements OnInit, OnDestroy {
     this.currentProject = this._multipleProjects.currentProject()?.name || '';
   }
 
-  private handleSelect(selectedId: string): void {
+  private handleSelect(item: any): void {
```
neogucky, line 75: any

projects/step-core/src/lib/modules/entity/types/select-entity-context.interface.ts
```diff
@@ -2,6 +2,6 @@ import { LegacyTableHandle } from '../../table/table.module';
 
 export interface SelectEntityContext extends LegacyTableHandle {
   readonly multipleSelection?: boolean;
-  handleSelect?(selectedId: string): void;
+  handleSelect?(selectedId: any): void;
```
neogucky, line 5: any

projects/step-frontend/src/lib/modules/timeseries/performance-view/filter-bar/item/entity-search-value.ts
```diff
@@ -0,0 +1,4 @@
+export interface EntitySearchValue {
+  searchValue: string;
+  entity?: any; // for labeling and displaying purposes
```
neogucky, line 3: any

neogucky (changes requested):
Looks good except for the many "any" types

projects/step-core/src/lib/modules/entity/types/select-entity-context.interface.ts
lucianbaciu: the generic select component we used to have, was based on id, which was always string. I changed it to allow the initiator to receive the entire entity that was selected (like a select). Since it's generic, I don't know how to specify any time

lucianbaciu:
@neogucky before you're ready to merge this, let me know, so I can create the PR for EE (otherwise it will not build anymore), so we can merge that fast afterwards

neogucky (dismissed):

projects/step-core/src/lib/services/function-actions-impl.service.ts
```diff
@@ -138,8 +138,8 @@ export class FunctionActionsImplService implements FunctionActionsService {
   selectFunction(): Observable<Keyword> {
     const selectedEntity$ = this._entityDialogs.selectEntityOfType('function', true);
     const function$ = selectedEntity$.pipe(
-      map((result) => result.item),
-      switchMap((id) => this._functionApiService.getFunctionById(id))
+      map((result) => result.item as Keyword),
+      switchMap((f: Keyword) => this._functionApiService.getFunctionById(f.id!))
```
dvladir, line 142: More readable. Typescript will resolve the type automatically, because there is cast at the previous line
```suggestion
      switchMap((keyword) => this._functionApiService.getFunctionById(keyword.id!))
```

projects/step-core/src/lib/services/plan-dialogs.service.ts
```diff
@@ -40,7 +40,7 @@ export class PlanDialogsService implements PlanLinkDialogService {
   selectPlan(): Observable<Plan> {
     const selectedEntity$ = this._entityDialogs.selectEntityOfType('plans', true);
     const plan$ = selectedEntity$.pipe(
-      map((result) => result.item),
+      map((result) => (result.item as Plan).id!),
       switchMap((id) => this._plansApiService.getPlanById(id))
```
dvladir, line 44: What about to do this in the same way for keywords?
```suggestion
      map((result) => result.item as Plan),
      switchMap((plan) => this._plansApiService.getPlanById(plan.id!))
```

projects/step-frontend/src/lib/modules/execution/components/execution-picker/execution-picker.component.ts
```diff
@@ -0,0 +1,63 @@
...
+  selector: 'step-execution-list',
+  templateUrl: './execution-picker.component.html',
+  styleUrls: ['./execution-picker.component.scss'],
+  providers: [
+    tablePersistenceConfigProvider('executionList', STORE_ALL),
+    selectionCollectionProvider<string, ExecutiontTaskParameters>('id', AutoDeselectStrategy.DESELECT_ON_UNREGISTER),
+    {
+      provide: BulkOperationsInvokeService,
+      useClass: ExecutionBulkOperationsInvokeService,
+    },
```
dvladir, line 31: Sync with master and change bulk operations configuration

projects/step-frontend/src/lib/modules/timeseries/pages/execution-page/ts-execution-page.component.ts
```diff
@@ -118,33 +147,8 @@ export class ExecutionPerformanceComponent implements OnInit, OnDestroy, OnChang
...
-        },
-        {
-          label: 'Name',
-          attributeName: 'name',
-          type: FilterBarItemType.FREE_TEXT,
-          isLocked: true,
-        },
-      ],
+      activeFilters: pageFilters,
+      filterOptions: JSON.parse(JSON.stringify(pageFilters)),
```
dvladir, line 151: In case if you want to have a copy of the default object, it make sense to create a separate method, that will construct and return filters.
```
activeFilters: this.createFilters(),
filterOptions: this.createFilters()
```
Both fields will have similar values, but with completely different references.
I think it's better than parse/stringify invocation

dvladir (changes requested):

dvladir (approved):

neogucky (approved):

PR #578, opened 2023-09-12
SED-2401 hide unsupported resource type on new resource dialog

neogucky:


projects/step-core/src/lib/components/predefined-options-input/predefined-options-input.component.html
```diff
@@ -4,7 +4,8 @@
   <input
     [ngModel]="value"
     (ngModelChange)="onValueChange($event)"
-    [disabled]="isDisabled"
+    [disabled]="true"
+    [class.hide-disabled]="!isDisabled"
```
dvladir, line 8: If the idea was to make the input non editable, and make it grey only for the disabled state, I think this is the better variant
```suggestion
    [disabled]="isDisabled"
    [readonly]="true"
```

projects/step-core/src/lib/components/predefined-options-input/predefined-options-input.component.scss
```diff
@@ -5,6 +5,10 @@ button.mat-menu-item {
   line-height: 1;
 }
 
+input.hide-disabled:disabled {
+  background: white;
+}
```
dvladir, line 10: @neogucky  In case if you will apply the readonly attribute, this can be removed

neogucky: Thanks, makes sense!

PR #600, opened 2023-10-09
SED-2468 [OS] Import dialog migrated

kmaslesa:
image


projects/step-core/src/lib/components/import-dialog/import-dialog.component.ts
```diff
@@ -0,0 +1,60 @@
...
+            }
+          },
+          complete: () => {
+            this.isImporting = false;
+          },
+          error: (error: any) => console.error(error),
+        });
+    } else {
+      this._dialogs.showErrorMsg('Upload not completed.');
+    }
```
dvladir, line 52: It will be easier to read, in case if the amount of nested blocks will be reduced.
Also try to avoid `any` usage

```suggestion
    if (!this.resourcePath) {
        this._dialogs.showErrorMsg('Upload not completed.');       
        return;
    }
 
    this.invokeImport()
      .pipe(
        tap(() => {
          this.isImporting = true;
        })
      )
      .subscribe({
        next: (response: string[]) => {
          this._matDialogRef.close();
           if (!!response?.length) {
            this._dialogs.showListOfMsgs(response);
          }
        },
        complete: () => {
          this.isImporting = false;
        },
        error: (error: unknown) => console.error(error),
       });
```

dvladir (changes requested):

projects/step-core/src/lib/components/import-dialog/import-dialog.component.scss
```diff
@@ -0,0 +1,18 @@
+div[step-modal-window-body],
+.advanced-body {
+  display: flex;
+  flex-direction: column;
+  gap: 1.5rem;
+
+  div {
+    max-width: 800px;
```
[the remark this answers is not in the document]
kmaslesa: sure, thanks!

projects/step-core/src/lib/components/import-dialog/import-dialog.component.ts
kmaslesa: thanks!

projects/step-core/src/lib/components/import-dialog/import-dialog.component.ts
```diff
@@ -0,0 +1,59 @@
+import { Observable, tap } from 'rxjs';
+import { Component, HostListener, inject } from '@angular/core';
+import { MAT_DIALOG_DATA, MatDialogRef } from '@angular/material/dialog';
+import { AlertType, DialogsService, ImportsService } from '@exense/step-core';
```
dvladir, line 4: Circular imports

dvladir (changes requested):

dvladir (approved):

neogucky (approved):
Good work thanks!

PR #606, opened 2023-10-11
SED-2483 [OS] Grid quota manager migrated

kmaslesa:
image


projects/step-frontend/src/lib/modules/grid/components/quota-manager/quota-manager.component.ts
```diff
@@ -0,0 +1,51 @@
...
+  ngOnInit(): void {
+    this.getQuotaManagerStatus();
+  }
+
+  getQuotaManagerStatus(): void {
+    this._augmentedQuotaManagerService.getQuotaManagerStatus().subscribe({
+      next: (response: string) => {
+        if (response.includes(':')) {
+          this.statusTitle = response.trim();
+          this.statusBody = '';
```
neogucky, line 25: We should do the same as for the case with "." here. Since copying would duplicate code, there should be a function:
private renderMessage(delimiter: string) : {title: string, body: string}

neogucky (changes requested):
Please also render the message if we have a quota

neogucky (approved):

PR #608, opened 2023-10-12
SED-2482 migrate admin settings

dvladir:


projects/step-core/src/lib/services/view-registry.service.ts
```diff
@@ -146,15 +153,99 @@ export class ViewRegistryService {
...
+        parentChildren = [];
+        this.temporaryRouteChildren.set(parentPathPart, parentChildren);
+      }
+    }
+
+    return parentChildren;
+  }
+
+  registerRoute(route: Route, { parentPath, label, weight }: SubRouterConfig = {}): void {
+    const root = this._router.config.find((route) => route.path === 'root');
```
neogucky, line 206: I would prefer to move all parameters into one Object, probably the best way would be by extending Route to include parentPath, label, weight

neogucky (changes requested):
Looks good, please check my comment for a small change

projects/step-core/src/lib/services/view-registry.service.ts
dvladir: @neogucky 
I don't think, that it would be correct to unite in one object.
`route` object can contain the whole subtrees, adding `parentPath` field to it, would bring some kind of duality in route declaration - that we can define both parent and children.
In our case `parentPath` is required only for those cases, when plugin extends a view.
About `label` and `weight`, if you inspect the code further you will see, that these fields are merged in route's data object.
So in other words this `SubRouteConfig` is just a syntax sugar to simplify the route's registration

neogucky (approved):

PR #614, opened 2023-10-23
SED-2407 ability to display different metric types in the charts

lucianbaciu:


projects/step-frontend/src/lib/modules/timeseries/chart/tooltip-plugin.ts
```diff
@@ -96,13 +96,15 @@ export class TooltipPlugin {
...
-          let allSeriesLength = yPoints.length;
-          let elementsToSelect = 5;
+          const allSeriesLength = yPoints.length;
+          const elementsToSelect = 5;
           let elipsisBefore = true;
           let elipsisAfter = true;
-          if (yPoints.length > elementsToSelect) {
-            var closestIndex = this.getClosestIndex(hoveredValue, yPoints);
+          const closestIndex = this.getClosestIndex(hoveredValue, yPoints);
+          if (closestIndex > -1) {
```
neogucky, line 104: > -1 sounds like you might mean:
```suggestion
          if (closestIndex >= 0) {
```

projects/step-frontend/src/lib/modules/timeseries/metric-chart/metric-chart.component.ts
```diff
@@ -0,0 +1,232 @@
...
+  @Input() settings!: MetricType;
+  @Input() range!: TSTimeRange;
+  @Input() allowGroupingChange = true;
+  @Input() allowMetricChange = true;
+  @Input() colorsPool: TimeseriesColorsPool = new TimeseriesColorsPool();
+
+  readonly AGGREGATES: AggregationType[] = ['SUM', 'AVG', 'MIN', 'MAX', 'COUNT', 'RATE', 'MEDIAN', 'PERCENTILE'];
+  selectedAggregate!: AggregationType;
+  groupingAttributes: MetricAttributeSelection[] = [];
+  selectedRange!: TSTimeRange; // when zooming, this will get smaller
```
neogucky, line 49: Should work like this:

```suggestion
  selectedRange = this.range;
```

projects/step-frontend/src/lib/modules/timeseries/metric-chart/metric-chart.component.ts
```diff
@@ -0,0 +1,232 @@
...
+
+  private _timeSeriesService = inject(TimeSeriesService);
+
+  ngOnInit(): void {
+    if (!this.settings) {
+      throw new Error('Metric settings input is mandatory');
+    }
+    if (!this.range) {
+      throw new Error('Range input is mandatory');
+    }
```
neogucky, line 61: These check should not be in ngOnInit but in ngOnChanges:

`ngOnChanges(changes: SimpleChanges): void {
    const cSettings = changes['settings'];
    if (!cSettings?.currentValue) { throw .... }
`

projects/step-frontend/src/lib/modules/timeseries/time-series.config.ts
```diff
@@ -71,6 +71,45 @@ export class TimeSeriesConfig {
...
+      if (num && num < 10) {
+        return num.toString();
+      }
+      const lookup = [
+        { value: 1, symbol: '' },
+        { value: 1e3, symbol: 'k' },
+        { value: 1e6, symbol: 'M' },
+        { value: 1e9, symbol: 'B' },
+      ];
+      const rx = /\.0+$|(\.[0-9]*[1-9])0+$/;
```
neogucky, line 86: Always try to select useful function names and avoid abreviations when possible. From the RegEx I would guess this matches numbers ending in 0 after the decimal point, if so I would name it "matchDecimalEndingZero" or something in that direction.

Ideally the lookup variable would also be a bit more self explaining.

projects/step-frontend/src/lib/modules/timeseries/time-series.config.ts
```diff
@@ -71,6 +71,45 @@ export class TimeSeriesConfig {
...
+    },
+    time: (milliseconds: number) => {
+      const lookup = [
+        { value: 1, symbol: ' ms' }, // milliseconds
+        { value: 1000, symbol: ' s' }, // seconds
+        { value: 60 * 1000, symbol: ' m' }, // minutes
+        { value: 3600 * 1000, symbol: ' h' }, // hours
+        // Add more units if needed, like { value: 86400 * 1000, symbol: 'd' } for days
+      ];
+      const rx = /\.0+$|(\.[0-9]*[1-9])0+$/;
```
neogucky, line 103: seems duplicated

projects/step-frontend/src/lib/modules/timeseries/time-series.config.ts
```diff
@@ -71,6 +71,45 @@ export class TimeSeriesConfig {
...
+      var item = lookup
+        .slice()
+        .reverse()
+        .find(function (item) {
+          return milliseconds >= item.value;
+        });
```
neogucky, line 109: Also looks duplicated. I would extract this as a function

projects/step-frontend/src/lib/modules/timeseries/util/timeseries-colors-pool.ts
```diff
@@ -17,6 +17,8 @@ export class TimeseriesColorsPool {
    * @param key - If it is not new, the existing value will be returned, otherwise a new color is generated.
    */
   assignColor(key: string): string {
```
neogucky, line 19: There is redundant code:

`
assignColor(key: string): string {
    let color = this.randomRGBA();
    if (this.predefinedColors.length !== 0) {
      color = this.predefinedColors[0];
      this.predefinedColors.shift();
    }
    this.assignedColors[key] = color;
    return color;
  }
`

projects/step-frontend/src/lib/modules/timeseries/util/timeseries-colors-pool.ts
```diff
@@ -41,11 +43,28 @@ export class TimeseriesColorsPool {
...
-    let alpha = Math.max(Number(r().toFixed(1)), 0.5);
-    return 'rgba(' + o(r() * s) + ',' + o(r() * s) + ',' + o(r() * s) + ',' + alpha + ')';
+
+    // Convert a number to a 2-digit hexadecimal string
+    const toHex = (num: number) => num.toString(16).padStart(2, '0');
+
+    // Generate RGB values
+    const red = o(r() * s);
+    const green = o(r() * s);
+    const blue = o(r() * s);
```
neogucky, line 65: I would prefer real function names:

const red = Math.round(Math.random() * 255);

neogucky (changes requested):
Some small fixes required

projects/step-frontend/src/lib/modules/timeseries/metric-chart/metric-chart.component.ts
lucianbaciu: @neogucky agree. I just checked the documentation and indeed the onChanges is called before onInit in the first place. I wasn't sure on this

projects/step-frontend/src/lib/modules/timeseries/util/timeseries-colors-pool.ts
lucianbaciu: @neogucky I don't think this is redundant. The randomRGBA() would be called all the time, even though 90% of the time it goes on the second condition. I don't like to have functions called if they are not needed, even if the complexity of the function is not very big...

projects/step-frontend/src/lib/modules/timeseries/time-series.config.ts
lucianbaciu: I did simplify it, but I would not group them into methods too much. Currently they have similar mapping, but I won't be surprised if we want change the formatting for one or the other only.

neogucky (approved):

PR #624, opened 2023-11-06
SED-2545 Opening a plan from another plan editor menu is not working

dvladir:


lucianbaciu (changes requested):

projects/step-core/src/lib/services/plan-dialogs.service.ts
```diff
@@ -118,7 +119,11 @@ export class PlanDialogsService implements PlanLinkDialogService {
   private openPlanInternal(planEditLink: string, artefactId?: string): void {
     const queryParams = artefactId ? { [ARTEFACT_ID]: artefactId } : undefined;
     const commands = planEditLink.split('/');
-    this._router.navigate(commands, { queryParams });
+    if (!this._router.url.includes(EDITOR_URL)) {
+      this._router.navigate(commands, { queryParams });
+      return;
+    }
+    from(this._router.navigateByUrl('/')).subscribe(() => this._router.navigate(commands, { queryParams }));
```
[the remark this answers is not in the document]
dvladir: @lucianbaciu 
`skipLocationChange` adds another bug. On the back button, the location was changed but UI hasn't been updated.
So in case of plan editor, I solve the issue in another way.

Please note, that I have to bring back two legacy templates, that have been removed during the clean-up, because they are actually using.

projects/step-core/src/lib/components/link-button/link-button.component.ts
```diff
@@ -28,6 +28,6 @@ export abstract class LinkButtonComponent implements OnInit {
 
     from(this._router.navigateByUrl('/'))
       .pipe(switchMap(() => timer(100)))
-      .subscribe(() => this._router.navigateByUrl(this.url!));
+      .subscribe(() => this._router.navigateByUrl(this.url!, { replaceUrl: true }));
```
lucianbaciu, line 31: what's the purpose of this replaceUrl? also, do we need the timer() thing?

lucianbaciu (approved):
I did some testing on my end too, and it seems that subscribing to ActivatedRoute is the only way to handle such scenarios in angular.

projects/step-core/src/lib/components/link-button/link-button.component.ts
dvladir: @lucianbaciu 
I've rechecked it and removed the timer. It's unnecessary now.
It was added to handle some issues when both angular & angularJS routing were used on the settings screen.
Now as this part it migrated, timer not needed anymore here.

`replaceUrl` parameter overrides last location change. So during the back navigation `this._router.navigateByUrl('/')` will be ignored. Nearly the same operation like `skipLocationChange`

PR #625, opened 2023-11-06
SED-2566 Rename scheduler tasks to schedules in the UI

dvladir:


projects/step-core/src/lib/client/table/services/table-api-wrapper.service.ts
```diff
@@ -1,19 +1,32 @@
...
+    pollAsyncTask(this._asyncTaskService),
+    map((status: AsyncTaskStatusResource) => {
+      if (!status?.result?.id) {
+        throw 'Invalid attachment id';
+      }
+      return status.result.id;
+    }),
+    tap((attachmentId) => this.downloadDatasource(attachmentId))
+  );
+
```
jeromecomte, line 29: Doesn't seem to be related to the ticket but I assume it makes sense

projects/step-frontend/src/lib/modules/scheduler/components/scheduler-configuration/scheduler-configuration.component.html
```diff
@@ -2,10 +2,10 @@
   <mat-card-content>
     <step-form-field>
       <step-label>
-        Enable scheduler
```
jeromecomte, line 5: This should be kept unchanged. This is still the scheduler who is enabled / disabled

projects/step-frontend/src/lib/modules/scheduler/components/scheduler-configuration/scheduler-configuration.component.html
```diff
@@ -2,10 +2,10 @@
...
+        Enable schedules
         <step-icon
           name="help-circle"
-          matTooltip="Used to enable or disable the scheduler globally accross all projects. Manual task execution remains possible."
+          matTooltip="Used to enable or disable the schedules globally accross all projects. Manual schedule execution remains possible."
```
jeromecomte, line 8: Keep "scheduler" and thus change to "Used to enable or disable the scheduler globally accross all projects. Manual schedule execution remains possible."

projects/step-frontend/src/lib/modules/scheduler/scheduler.module.ts
```diff
@@ -68,7 +68,7 @@ export class SchedulerModule {
         },
         {
           parentPath,
-          label: 'Scheduler',
+          label: 'Schedules',
```
jeromecomte, line 71: Keep "Scheduler" as the settings relate to the Scheduler

jeromecomte (dismissed):
Minor changes request

projects/step-core/src/lib/client/table/services/table-api-wrapper.service.ts
dvladir: @jeromecomte During this task implementation, I found out that monitoring screen didn't work because it referenced to legacy export service from angularJs, that was cleaned up from the project.
I fixed it by adding necessary method to this service

neogucky (approved):

PR #636, opened 2023-11-13
SED-2485 dialogs service migration

bojidaryovchev:
https://exense.atlassian.net/browse/SED-2485


projects/step-core/src/lib/directives/dialog-close.directive.ts
```diff
@@ -0,0 +1,14 @@
+import { Directive, HostListener, inject } from '@angular/core';
```
dvladir, line 1: This directive copies `mat-dialog-close` functionality.
Remove his directive and replace it's usages with `mat-dialog-close`

dvladir (changes requested):

projects/step-core/src/lib/directives/dialog-close.directive.ts
bojidaryovchev: this is for a reason, the mat-dialog-close directive closes the dialog with an empty string instead of undefined (undefined is returned when the dialog is closed by clicking outside) and that is first of all inconsistent and second its incorrect in my opinion for every dialog to return an empty string (for example we have this dialog to edit text and it doesnt make sense for it to return an empty string upon cancel, so we want the dialog to close with undefined which we can check against and be sure that the string that gets emitted is the actual value the user intended)

bojidaryovchev: https://github.com/exense/step-frontend/pull/636/files#diff-bc3e4f53d59cb8ba11bed91ac8a33b73f0d0c07e9bbe66957b1a9e2ba5513d8dR47

bojidaryovchev: in `dialogs.service.ts`
image

dvladir: @bojidaryovchev 
In case if you need the close result other than empty string, use:
`[matDialogClose]="undefined"`

bojidaryovchev: ah, spot on, my bad, didnt think to check the docs -_-

bojidaryovchev:
useful remarks

dvladir (dismissed):

bojidaryovchev:
added `exports.js` contents back as part of this PR as several things are broken due to it, should I revert that or? @dvladir

dvladir:
> added `exports.js` contents back as part of this PR as several things are broken due to it, should I revert that or? @dvladir

@bojidaryovchev Yes, please. Synchronize with master branch and remove legacy export service from this PR. It's no longer needed

neogucky (approved):

PR #637, opened 2023-11-13
SED-2581 New views to list and define alerting rules

dvladir:


david-stephan (approved):

projects/step-core/src/lib/modules/basics/components/string-array-input/string-array-input.component.ts
```diff
@@ -0,0 +1,77 @@
...
+  @Input() label?: string;
+  @Input() showRequiredMarker: boolean = false;
+  @Input() tooltip?: string;
+  @Input() errorsDictionary?: Record<string, string>;
+
+  private terminator$ = new Subject<void>();
+  private _fb = inject(FormBuilder).nonNullable;
+
+  private onChange?: OnChange;
+  private onTouch?: OnTouch;
```
neogucky, line 24: Why are we using event based architecture instead of observable?

dvladir: @neogucky It's a part of ControlValueAccessor implementation, which is used to make component work with angular's forms (ngModel, formControl, formControlName)

neogucky (dismissed):

neogucky (dismissed):

neogucky (approved):

PR #638, opened 2023-11-16
SED-2588-Custom charts display entities names instead of ids

lucianbaciu:


lucianbaciu:
Task link: https://exense.atlassian.net/browse/SED-2588
The task has a ss in it.

projects/step-frontend/src/lib/modules/timeseries/metric-chart/metric-chart.component.ts
```diff
@@ -177,6 +219,45 @@ export class MetricChartComponent implements OnInit, OnChanges {
...
+   */
+  private fetchAndSetLabels<T>(
+    series: TSChartSeries[],
+    groupDimensionIndex: number,
+    fetchByIds: (ids: string[]) => Observable<T[]>,
+    mapEntityToLabel: (entity: T) => string | undefined
+  ) {
+    const ids: string[] = series.map((s) => s.labelItems[groupDimensionIndex] as string).filter((id) => id);
+
+    if (ids.length > 0) {
```
dvladir, line 238: I would prefer to invert the condition and do a return, instead of adding additional nesting block
```
if (!ids.length) {
  return;
}
```

projects/step-frontend/src/lib/modules/timeseries/metric-chart/metric-chart.component.ts
```diff
@@ -177,6 +219,45 @@ export class MetricChartComponent implements OnInit, OnChanges {
...
+        series.forEach((s) => {
+          const entityId = s.labelItems[groupDimensionIndex];
+          if (entityId) {
+            const entity = entitiesByIds.get(entityId);
+            if (entity) {
+              this.chart.setLabelItem(s.id, groupDimensionIndex, mapEntityToLabel(entity));
+            }
+          }
+        });
+      });
```
dvladir, line 257: In case of Observables it make sense to do all required transformations inside the pipe chain,
and then do the final action inside the subscribe.

```suggestion
      fetchByIds(ids)
          .pipe(
              map((entities) => entities.reduce((res, entity) => {
                  const entityId = (entity as any).id as string;
                  if (entityId) {
                     res[entityId] = entity;
                  }
                  return res;
              }, {} as Record<string, T>)),
              map((entitiesDictionary) => series
                    .map(({id, labelItems}) => {
                         const entity = entitiesDictionary[labelItems[groupDimensionIndex]];
                         const entityLabel = entity ? mapEntityToLabel(entity) : undefined;
                         return {id, entityLabel};
                    })
                    .filter((seriesLabels) => !!seriesLabels.entityLabel)
              )
          )
          .subscribe((seriesLabels) => { 
               seriesLabels.forEach(({id, entityLabel}) => this.chart.setLabelItem(id, groupDimensionIndes, entityLabel);
          });
```

dvladir (changes requested):

projects/step-frontend/src/lib/modules/timeseries/metric-chart/metric-chart.component.ts
lucianbaciu: @dvladir this looks to me as over-engineering . I see no reason to do it: it just adds complexity - both visually and technically. It's my code and I barely follow what the code is doing. If you consider it is a need though, I will add it.

jeromecomte (changes requested):
See my important remark

projects/step-frontend/src/lib/modules/timeseries/metric-chart/metric-chart.component.ts
dvladir: @lucianbaciu In concrete this case, both variants will work. I would say that this remark is not strictly required.
But in common, from the point of view how we work with observable streams, it make sense to do the transformations in pipe chains, to receive another stream with required data.
Then we can decide, will we subscribe to this new stream, will we reuse it as part of another stream or even will we pass the new stream as parameter somewhere else.
In other words wrapping data in the observable streams, increases the abstraction level

dvladir (approved):

projects/step-frontend/src/lib/modules/timeseries/metric-chart/metric-chart.component.ts
```diff
@@ -177,6 +219,45 @@ export class MetricChartComponent implements OnInit, OnChanges {
...
+    groupDimensionIndex: number,
+    fetchByIds: (ids: string[]) => Observable<T[]>,
+    mapEntityToLabel: (entity: T) => string | undefined
+  ) {
+    const ids: string[] = series.map((s) => s.labelItems[groupDimensionIndex] as string).filter((id) => id);
+
+    if (ids.length > 0) {
+      fetchByIds(ids).subscribe((entities) => {
+        const entitiesByIds = new Map<string, T>();
+        entities.forEach((entity) => {
```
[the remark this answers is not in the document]
lucianbaciu: @jeromecomte it doesn't. the order of the elements does not matter

projects/step-frontend/src/lib/modules/timeseries/metric-chart/metric-chart.component.ts
lucianbaciu: @dvladir thank you. I know and I agree. If there would have been another other async task, or chains, I would definitely go for your suggestion.

bojidaryovchev (approved):

PR #639, opened 2023-11-16
SED-2590-Locked tooltip will disable charts refresh

lucianbaciu:


projects/step-frontend/src/lib/modules/timeseries/chart/time-series-chart.component.ts
```diff
@@ -41,6 +41,7 @@ export class TimeSeriesChartComponent implements OnInit, AfterViewInit, OnChange
 
   @Output() onZoomReset = new EventEmitter<void>();
   @Output() onZoomChange = new EventEmitter<TSTimeRange>();
+  @Output() onLockStateChange = new EventEmitter<boolean>();
```
dvladir, line 44: https://angular.io/guide/styleguide#style-05-16
Don't name events with `on` prefix.
Linter should warned about it. I didn't configured it as error, because unfortunately there are several places, where we such names still exists. But we should avoid of usage of such event names

projects/step-frontend/src/lib/modules/timeseries/chart/tooltip-plugin.ts
```diff
@@ -169,6 +169,12 @@ export class TooltipPlugin {
           bound = over;
           //	bound = document.body;
 
+          over.onclick = () => {
+            // @ts-ignore
+            const lockState = u.cursor._lock;
+            ref.onLockStateChange.next(lockState);
```
dvladir, line 175: Even `EventEmitter` is the `Subject`'s descendant, it make sense to follow the `EventEmitter` api and call `emit` instead of `next`, even both methods are doing the same

dvladir (changes requested):

dvladir (changes requested):

dvladir:
@lucianbaciu In the TimeSeriesContext I noticed the following:
```
 resetZoom() {
    if (JSON.stringify(this.selectedTimeRange) === JSON.stringify(this.fullTimeRange)) {
      // return; // this is causing some issues in the ranger. it's selection get 0 width, so better keep it like this for now.
    }
    }
    this.selectedTimeRange = this.fullTimeRange;
    this.selectedTimeRangeChange$.next(this.selectedTimeRange);
  }
```
If statement does nothing. Make sense to remove this block.

Also comparing two objects through the stringify isn't optimal. Especially it contains few fields:
```
isFullRangeSelected() {
    return JSON.stringify(this.selectedTimeRange) === JSON.stringify(this.fullTimeRange);
}
```
Please compare it explictly
```
isFullRangeSelected() {
    return this.selectedTimeRange.from === this.fullTimeRange.from &&
              this.selectedTimeRange.to === this.fullTimeRange.to;
}
```

projects/step-frontend/src/lib/modules/timeseries/time-series-context.ts
```diff
@@ -36,6 +36,7 @@ export class TimeSeriesContext {
   private readonly activeFilters$: BehaviorSubject<TsFilterItem[]>;
   private readonly filterSettings$: BehaviorSubject<TsFilteringSettings>;
   private readonly chartsResolution$: BehaviorSubject<number> = new BehaviorSubject<number>(0);
+  private readonly chartsLockedState$ = new BehaviorSubject<boolean>(false);
```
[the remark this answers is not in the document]
lucianbaciu: good catch

dvladir (approved):

PR #666, opened 2023-12-28
SED-2678 remove '#/root' from url

dvladir:


projects/step-core/src/lib/modules/basics/services/app-config-container.service.ts
```diff
@@ -24,6 +24,11 @@ export class AppConfigContainerService {
     if (forceClientUrl) {
       return this.defaultClientUrl;
     }
-    return this.conf?.defaultUrl ?? this.defaultClientUrl;
+    let result = this.conf?.defaultUrl ?? this.defaultClientUrl;
+    if (result.startsWith('/root')) {
+      //TODO: temporary fix, until server side default url isn't reconfigured
+      result = result.replace('/root', '');
+    }
```
neogucky, line 31: @dvladir can you remove this and check if it still works with the BE changes?

dvladir: Done

PR #667, opened 2023-12-28
SED-2423 TimeInput model/view measure calculation

dvladir:


projects/step-core/src/lib/modules/basics/components/time-input/operation-per-time-input.component.ts
```diff
@@ -0,0 +1,34 @@
...
+import { Component, ViewEncapsulation } from '@angular/core';
+import { NgControl } from '@angular/forms';
+
+@Component({
+  selector: 'step-operation-per-time-input',
+  templateUrl: './base-time-converter.component.html',
+  styleUrls: ['./base-time-converter.component.scss'],
+  encapsulation: ViewEncapsulation.None,
+})
+export class OperationPerTimeInputComponent extends BaseTimeConverterComponent {
```
neogucky, line 12: For me the naming doesn't make sense. I would name it "PerTimeUnitInputComponent" (etc.)

dvladir: Done

neogucky (approved):

PR #679, opened 2024-01-11
SED-2703 handle os plugins

dvladir:


projects/step-frontend/src/lib/os-plugins/plugin.module.ts
```diff
@@ -9,10 +13,10 @@ import { StepCoreModule, PluginLazyLoad, ImportMeta } from '@exense/step-core';
 export class PluginModule extends PluginLazyLoad {
```
neogucky, line 13: Why do we need export const registerOsPlugins = async (injector: Injector): Promise<void> => {... between this module and plugin-initializer? As we don't do any dynamic plugin loading in OS the code would be simpler to directly call this in the plugin-initializer.

dvladir: @neogucky dynamic import for PluginModule in OS has been removed

neogucky (approved):

PR #696, opened 2024-01-22
SED-2725 timeseries legend and filters display entities names

lucianbaciu:


neogucky (approved):
I think the hardcoded strings are ok in this case as it is only a small feature and otherwise would require two BE / FE enums

dvladir (dismissed):

neogucky (approved):

PR #698, opened 2024-01-23
SED-2774 Custom screen input for KW name to open the script editor

dvladir:


neogucky (approved):
Great refactoring of functionLink to go with this change!

PR #701, opened 2024-01-24
SED-2779 Generate YAML in UI plan editor

dvladir:


projects/step-frontend/src/lib/modules/plan-editor/components/plan-editor-actions/plan-editor-actions.component.ts
```diff
@@ -55,6 +56,7 @@ export class PlanEditorActionsComponent implements OnChanges {
     undo: 'Undo (Ctrl + Z)',
     redo: 'Redo (Ctrl + Y)',
     duplicate: 'Duplicate this plan',
+    showSource: `Show plan's YAML source`,
```
neogucky, line 59:
```suggestion
    showSource: 'Show plan's YAML source',
```

projects/step-frontend/src/lib/modules/plan-editor/injectables/plan-source-stub.service.ts
neogucky, line 1: @neogucky don't forget this has to be changed once the BE is ready!

neogucky (approved):
Small character issue & waiting for BE, else approved!

projects/step-frontend/src/lib/modules/plan-editor/components/plan-editor-actions/plan-editor-actions.component.ts
dvladir: @neogucky  \` quotes are explicitly used, because string already contains a single quote inside in word `plan's`

neogucky: I missed that, but in that case I would prefer to escape the quote instead of using another quote style

neogucky (approved):
Approved but would still like the quote change, leave it up to you

neogucky (approved):

PR #704, opened 2024-01-25
Sed 2747 step ts ranger not scaling correctly when resizing 

lucianbaciu:


lucianbaciu:
@neogucky final cause of the bug:

When we started development, we used uPlot **1.6.21.** When resizing, this version was not automatically adjust the selection (blue area), and I did it manually.
Since we both had node_modules already populated, it was not downloading a newer version. Even after I deleted my node_modules, it was fetching uPlot **1.6.26** (I don't know why), which does not get the fix yet.
In uPlot **1.6.28** we do have the auto adjust of the selection. With our existing code, it was doing it twice and accelerating the selection.

projects/step-frontend/src/lib/modules/timeseries/ranger/ts-ranger.component.ts
```diff
@@ -330,6 +315,7 @@ export class TSRangerComponent implements OnInit, AfterViewInit, OnChanges, OnDe
         ],
         setSelect: [
           (uRanger: uPlot) => {
+            console.log('set select called');
```
neogucky, line 318:
```suggestion
```

neogucky (approved):

PR #705, opened 2024-01-26
SED-2785 Connect Dialogs to Angular Routing

dvladir:


projects/step-core/src/lib/components/scheduler-task-link/scheduler-task-link.component.ts
```diff
@@ -22,6 +24,11 @@ export class SchedulerTaskLinkComponent implements CustomComponent {
     if (!this._logic || !this.context) {
       return;
     }
-    this._logic.editTask(this.context).subscribe();
+    // TODO temporary solution, while scheduler dialog supports two variants, how it can be opened
```
neogucky:
```suggestion
    // TODO temporary solution - scheduler dialog only uses new route based dialog when directly opend on the schedule page
```

PR #718, opened 2024-02-07
SED-2820 fix styles

lukasz-lepa:
https://exense.atlassian.net/browse/SED-2820


projects/step-core/src/lib/modules/automation-package-common/components/automation-package-filter-popover/automation-package-filter-popover.component.scss
```diff
@@ -12,6 +12,8 @@ mat-card {
   }
 
   mat-card-title {
-    font-size: 2rem;
+    font-size: 1.8rem;
+    padding: 16px;
```
neogucky, line 16:
```suggestion
    padding: 1.6rem;
```
We try to avoid px, 1rem = 10px

lukasz-lepa: Sure, changed px to rem

neogucky (approved):

PR #741, opened 2024-02-22
SED-2856 Remove pencil button from number inputs

dvladir:


projects/step-core/src/lib/modules/dynamic-forms/components/dynamic-textfield/dynamic-textfield.component.html
```diff
@@ -1,28 +1,29 @@
...
-  [showRequiredMarker]="showRequiredMarker"
-  [ngModel]="expression"
-  (ngModelChange)="onExpressionChange($event)"
-  [disabled]="isDisabled"
-  (toggleConstantValue)="toggleDynamicExpression(false)"
-  (blur)="onBlur()"
-></step-expression-input>
-
-<ng-template #constantValueTemplate>
+@if (dynamic) {
```
neogucky, line 1: I like the new @if directive! Looks cleaner than then the old else-template solution

neogucky (approved):

neogucky (approved):

PR #755, opened 2024-03-04
SED-2845 Make sure links are used for navigation

dvladir:


projects/step-core/src/lib/pipes/plan-url.pipe.ts
```diff
@@ -0,0 +1,19 @@
...
+
+@Pipe({
+  name: 'planUrl',
+})
+export class PlanUrlPipe implements PipeTransform {
+  static transform(idOrPlan?: string | Plan): string {
+    if (!idOrPlan) {
+      return '';
+    }
+    const id = typeof idOrPlan === 'string' ? idOrPlan : idOrPlan.id;
```
neogucky, line 12:
```suggestion
    const id = (typeof idOrPlan === 'string' ? idOrPlan : idOrPlan.id);
```

I prefer brackets for readability

projects/step-core/src/lib/pipes/plan-url.pipe.ts
```diff
@@ -0,0 +1,19 @@
...
+    return `/plans/editor/${id}`;
+  }
+
+  transform(idOrPlan?: string | Plan): string {
+    return PlanUrlPipe.transform(idOrPlan);
+  }
```
neogucky, line 18: Why do we need the static method? I would move the code in here.

dvladir: @neogucky 
Sometimes it's required to invoke transform logic directly from component or service.
If you don't like the static method, the following alternatives could be used:

1. Pipe can be provided as service on component's level and then injected. This won't work in case if transform invocation required from service, that is provided in root.
2. Transformation logic is moved to the service, than pipe invokes the service. I think it's more clear approach. Also, as we have to do the same for other entities, I thinks it make sense to create such service that will contain editor's urls transformation logic for all of them.

neogucky (approved):

neogucky (approved):

PR #764, opened 2024-03-14
SED-2901 Editing composite keywords needs more context information

dvladir:


projects/step-core/src/lib/modules/editable-labels/components/editable-component/editable.component.ts
```diff
@@ -33,19 +36,24 @@ type LabelTemplateAccessor = Mutable<Pick<EditableComponent<any>, 'labelTemplate
...
   protected _elementRef = inject<ElementRef<HTMLElement>>(ElementRef);
   protected _changeDetectorRef = inject(ChangeDetectorRef);
   protected _document = inject(DOCUMENT);
 
-  @Input('labelTemplate') labelTemplateInput!: TemplateRef<{}>;
+  /** @Input('labelTemplate') **/
+  labelTemplateInput = input<TemplateRef<{}> | undefined>(undefined, {
+    alias: 'labelTemplate',
+  });
+
```
neogucky, line 48: On first glance this code looks way more complicated than the old syntax. What is the advantage of this?

neogucky (dismissed):

projects/step-core/src/lib/modules/editable-labels/components/editable-component/editable.component.ts
dvladir: @neogucky In current case converting `labelTemplateInput` input and `labelTemplateDirective` content child to signals, easily allows to compute `labelTemplate` signal and remove `ngAfterContentInit` and `ngOnChanges` hooks.

neogucky (approved):

PR #767, opened 2024-03-15
SED-2888 remove footer, add icon before name

lukasz-lepa:
https://exense.atlassian.net/browse/SED-2888


dvladir (changes requested):
@lukasz-lepa Please, check comments in task's ticket

projects/step-core/src/lib/components/artefact-details/artefact-details.component.html
```diff
@@ -67,8 +78,5 @@
       </section>
     </mat-expansion-panel>
     <hr />
```
neogucky, line 80:
```suggestion
```

neogucky (approved):

PR #773, opened 2024-03-18
SED-2708-refresh feature + improvements

lucianbaciu:


projects/step-frontend/src/lib/modules/timeseries/components/dashboard/dashboard.component.ts
```diff
@@ -290,19 +335,35 @@ export class DashboardComponent implements OnInit, OnDestroy {
...
+              dashlets$.push(this.refreshRanger());
+            }
+            return forkJoin(dashlets$).pipe(defaultIfEmpty([]));
+          }),
+          tap(() => (this.refreshInProgress = false)),
+        )
+        .subscribe();
+    } else {
+      // skip refresh
+    }
```
dvladir, line 366: Instead of empty else block, I would prefer to check the condition at the beginning and exit. It will reduce the code's block nesting and will make it more readable.

```suggestion
    if (this.refreshInProgress && !force) {
        // skip refresh
        return;
    }

    this.refreshSubscription?.unsubscribe();
    this.refreshInProgress = true;
    this.refreshSubscription = of(null)
      .pipe(
        tap(() => (this.refreshInProgress = true)),
        switchMap(() => {
          const dashlets$ = this.dashlets?.map((dashlet) => dashlet.refresh());
          if (refreshRanger) {
            dashlets$.push(this.refreshRanger());
          }
          return forkJoin(dashlets$).pipe(defaultIfEmpty([]));
        }),
        tap(() => (this.refreshInProgress = false)),
      )
      .subscribe();
```

dvladir (changes requested):

neogucky (changes requested):
Looks good to me, after Vladimirs point is improved we can merge

neogucky (dismissed):

neogucky (approved):

PR #778, opened 2024-03-25
SED-2978 POC: user avatars

dvladir:


neogucky (dismissed):

projects/step-frontend/src/lib/components/main-view/main-view.component.scss
```diff
@@ -1,3 +1,13 @@
 #mainView {
   height: 100%;
 }
+
+.user {
```
lucianbaciu, line 5: ambigous class name. I'd rather use "user-label", "user-section" or something similar depending on what you display in current-user-avatar

projects/step-frontend/src/lib/modules/admin/components/user-avatar/user-avatar.component.ts
```diff
@@ -0,0 +1,35 @@
...
+  );
+
+  protected userName$ = this.user$.pipe(map((user) => user?.username ?? this.userId()));
+
+  protected color$ = this.user$.pipe(
+    map((user) => {
+      if (!this._auth.getConf()?.authentication) {
+        return ANONYMOUS_COLOR;
+      }
+      return user?.preferences?.preferences?.[this._avatarColorPreferenceKey];
```
lucianbaciu, line 32: some model issues on ".preferences.preferences"?
maybe should return a default color in case something is undefined, or it's not the case?

lukasz-lepa (dismissed):

neogucky (dismissed):

PR #788, opened 2024-04-04
SED-3002 Support agent autoscaling in FE

dvladir:


projects/step-frontend/src/lib/modules/execution/components/execution-progress/execution-progress.component.html
```diff
@@ -14,6 +14,31 @@
...
+      @if (
+        autoscalingStatus?.statusDescription ||
+        autoscalingStatus?.tokenCountStarted ||
+        autoscalingStatus?.tokenCountStarted
+      ) {
+        <step-alert [type]="AlertType.INFO" class="auto-scaling-status">
+          <div>Auto scaling status</div>
+          <div>{{ autoscalingStatus?.statusDescription }}</div>
+          @if (autoscalingStatus?.tokenCountStarted) {
+            <div>Token count started</div>
```
jeromecomte, line 26: Would prefer to display something like: "Progress: x/y agent token started" where x = tokenCountStarted and y = tokenCountTarget

projects/step-core/src/lib/client/generated/index.ts
```diff
@@ -18,6 +18,7 @@ export type { Attachment } from './models/Attachment';
 export type { AttachmentMeta } from './models/AttachmentMeta';
 export type { AttributeStats } from './models/AttributeStats';
 export type { AutomationPackage } from './models/AutomationPackage';
+export type { AutomationPackageUpdateResult } from './models/AutomationPackageUpdateResult';
```
jeromecomte, line 21: @dvladir This is not related to the BE changes performed. Do you know why we have this change in the generated client? The BE branch is based on master. I might have to rebase it on master?

jeromecomte (commented):
@dvladir PR looks OK. See my remarks/question. We should keep the PR open as long as all the subtickets of SED-2963 are not completed and merge everything to master at the same time

projects/step-core/src/lib/client/generated/index.ts
jeromecomte: @dvladir thanks for the clarification

neogucky (approved):

neogucky (approved):

PR #805, opened 2024-04-17
SED-3029 make input take whole space

lukasz-lepa:
https://exense.atlassian.net/browse/SED-3029


projects/step-core/src/lib/modules/custom-forms/components/custom-form/custom-form.component.scss
```diff
@@ -9,4 +9,8 @@
   &:not(.inline) {
     flex-direction: column;
   }
+
+  .flex-1 {
+    flex: 1;
+  }
```
dvladir, line 15: - It make sense for inline forms only.
- Avoid class names, that duplicates style rules

```suggestion
  .inline.fill-whole-space {
    flex: 1;
  }
```

dvladir (changes requested):

neogucky (dismissed):

dvladir (dismissed):

neogucky (approved):

PR #809, opened 2024-04-22
SED-2976 Analytics_widget_integration_into_any_view

neogucky:


projects/step-frontend/src/lib/modules/execution/components/execution-progress/execution-progress.component.ts
```diff
@@ -297,6 +303,65 @@ export class ExecutionProgressComponent
...
+       *  [aggregation]="AVG"
+       *  [unit]="ms"
+       *  [displayType]="line"
+       *  [renderingSettings]="..."
+       * />
+       *
+       */
+
+      this.chartContext = new TimeSeriesContext({
+        id: '123',
```
neogucky, line 339: what exactly does this ID reference? Does it make sense to manually reference something in the analytics view? What would happen if we use an ID multiple times?

projects/step-frontend/src/lib/modules/execution/components/execution-progress/execution-progress.component.ts
```diff
@@ -297,6 +303,65 @@ export class ExecutionProgressComponent
...
+        grouping: ['name'],
+        timeRange: { from: this.execution.startTime!, to: this.execution.endTime! },
```
neogucky, line 341: as we also want to show this for running executions the endTime will probably not work. As this is only an example we just have to fix it when moving it to the new execution view

projects/step-frontend/src/lib/modules/execution/components/execution-progress/execution-progress.component.ts
```diff
@@ -297,6 +303,65 @@ export class ExecutionProgressComponent
...
+      });
+
+      this.chartItem = {
+        name: 'Test chart with response times',
+        type: 'CHART',
+        size: 100,
```
neogucky, line 347: Is the size used for the rendering size? How are you adjusting that in the analytics view when resizing?

projects/step-frontend/src/lib/modules/execution/components/execution-progress/execution-progress.component.ts
lucianbaciu: @neogucky the Id is an internal identifier which is mostly used to identify a context (a dashboard). If you have 3 tabs with 3 executions open, you will have 3 contexts created. they will have the executionId as `id`. The id is also used to sync charts (or at least it was).

projects/step-frontend/src/lib/modules/execution/components/execution-progress/execution-progress.component.ts
lucianbaciu: @neogucky I remember that is only for the height

PR #811, opened 2024-04-24
SED-2892 custom dashboard   tooltip with execution links

lucianbaciu:


projects/step-frontend/src/lib/modules/timeseries/modules/chart/injectables/tooltip-plugin.ts
```diff
@@ -16,62 +16,13 @@ export class TooltipPlugin {
...
-    }
-    leftContainer.appendChild(nameDiv);
-    return leftContainer;
-  }
-
+  /**
+   * The tooltip will be displayed while the user will hover the chart.
+   * An execution is optional to display. It can be configured via settings and via chart metadata (where the actual
+   * data is taken from)
+   */
```
neogucky, line 23: This sounds more like infor for the KB than a needed comment in the code?

neogucky (approved):
Approved, but please check if the comments can't be reduced. Comments should only be used when really needed and be as concise as possible.

projects/step-frontend/src/lib/modules/timeseries/modules/chart/injectables/tooltip-plugin.ts
lucianbaciu: @neogucky it's for us. In order to have the execution link working there are a few things to configure - we had a bug for it and I had to investigate pretty much to remember why it is causing. It's important to mention that "settings and metadata" are needed for this.

lucianbaciu: I can simplify it though

neogucky (approved):

PR #824, opened 2024-05-13
SED-3073 add data testid

lukasz-lepa:
https://exense.atlassian.net/browse/SED-3073


projects/step-core/src/lib/directives/test-dataid.directive.ts
```diff
@@ -0,0 +1,14 @@
+import { Directive, ElementRef, inject, Input } from '@angular/core';
+
+@Directive({
+  selector: '[testDataId]',
```
dvladir, line 4: @neogucky @lukasz-lepa 
I think it make sense, to follow the naming convention and apply the selector with `step` prefix.
And make it like `[stepTestId]`

projects/step-core/src/lib/directives/test-dataid.directive.ts
```diff
@@ -0,0 +1,14 @@
...
+})
+export class TestDataIdDirective {
+  @Input('testDataId') testDataId: string | undefined;
+
+  private _el = inject(ElementRef);
+
+  ngOnInit() {
+    this._el.nativeElement.setAttribute('data-testid', this.testDataId);
```
dvladir, line 12: 1. `data-testid` name looks very common. I would also renamed it to `data-step-testid`, to specify explicitly that it is a **step** testing identifier.
2. It's recommended to use [Renderer](https://angular.io/api/core/Renderer2#description) service for manipulation with DOM

projects/step-core/src/lib/directives/test-dataid.directive.ts
```diff
@@ -0,0 +1,14 @@
+import { Directive, ElementRef, inject, Input } from '@angular/core';
+
+@Directive({
+  selector: '[testDataId]',
+})
+export class TestDataIdDirective {
+  @Input('testDataId') testDataId: string | undefined;
```
dvladir, line 7: I would add the behavior, that this identifier might change on the fly.
Even in most cases it will be a constant string, ability to react on this input change might prevent some future errors.
input signal + effect, instead of ngOnInit will look nice in this case

projects/step-core/src/lib/directives/test-dataid.directive.ts
```diff
@@ -0,0 +1,14 @@
...
+
+  private _el = inject(ElementRef);
```
dvladir, line 9: nativeElement type not specified.
```suggestion
  private _el = inject<ElementRef<HTMLElement>>(ElementRef);
```

dvladir (changes requested):

projects/step-core/src/lib/directives/test-dataid.directive.ts
neogucky: @lukasz-lepa @dvladir I discussed this and thought about it and finally agree, lets rename to data-step-testid

neogucky (commented):
Gernerally good work! Please check the remaining comments, then I will approve

projects/step-core/src/lib/directives/test-dataid.directive.ts
```diff
@@ -0,0 +1,20 @@
+import { Directive, effect, ElementRef, inject, input, Input, InputSignal, Renderer2 } from '@angular/core';
+
+@Directive({
+  selector: '[stepTestId]',
+})
+export class TestDataIdDirective {
+  testDataId: InputSignal<string | undefined> = input<string | undefined>();
```
dvladir, line 7: Currently now, directive selector and input attributed doesn't link with each other. The valid usage of such directive would be: `<input type="text" stepTestId testDataId="some-field"/>`, 
but we need to be it like: `<input type="text" stepTestId="some-field"/>`, 

To achieve it, provide an alias to the input field.
```suggestion
  testDataId: InputSignal<string | undefined> = input<string | undefined>({ alias: 'stepTestId' });
```

projects/step-frontend/src/lib/modules/admin/components/my-account/my-account.component.html
```diff
@@ -85,30 +93,40 @@
...
+            />
           </step-form-field>
         </td>
       </ng-container>
       <ng-container matColumnDef="actions">
         <th class="actions" mat-header-cell *matHeaderCellDef></th>
         <td mat-cell *matCellDef="let element">
-          <button mat-icon-button matTooltip="Delete preference" (click)="removePreference(element)">
+          <button
+            testDataId="delete-preference-btn"
```
dvladir, line 125: Please, also search for other place where `testDataId` has been applied, and change it to `stepTestId`
```suggestion
            stepTestId="delete-preference-btn"
```

dvladir (changes requested):

neogucky (approved):

dvladir (approved):

neogucky (approved):

PR #842, opened 2024-05-20
SED-3095 constraints on colors to avoid invisible series

lucianbaciu:


neogucky (approved):

projects/step-core/src/lib/modules/basics/components/marker/marker.component.scss
```diff
@@ -1,30 +1,56 @@
-.color-square {
+.marker-filled-square {
```
dvladir, line 1: As these css classes are used inside chart tooltip also, I think it make sense to move it from component to global application styles into `step-core/styles/component` as mixin and then include it in app's style.scss.

Angular adds component style-sheet, when component first renders on the page. So if component will never be used, it's stylesheet will be never included. It can cause the situation that chart's tooltip won't receive the required styles.

Also as these styles will become global, it makes sense to rename the classes as `.step-marker-filled-square`, `.step-marker-dots` and `.step-marker-dashed-square`

dvladir (changes requested):

dvladir (approved):

PR #890, opened 2024-07-10
SED-2959 persist analytics filters in singleton with reset button

lucianbaciu:


projects/step-frontend/src/lib/modules/timeseries/components/dashboard/dashboard-state-engine.ts
```diff
@@ -40,15 +41,38 @@ export class DashboardStateEngine {
...
+      case 'FULL':
+        if (timeRangeSettings.defaultFullRange?.from && !timeRangeSettings.defaultFullRange?.to) {
+          // to has to be set to now
+          timeRangeSettings.fullRange.to = new Date().getTime() - 5000;
+        }
+        break;
+      case 'ABSOLUTE':
+        break;
+      case 'RELATIVE':
+        const now = new Date().getTime() - 5000;
```
neogucky, line 55: I would move this over the switch and use it also in the 'FULL' case.

projects/step-frontend/src/lib/modules/timeseries/components/dashboard/dashboard-state-engine.ts
```diff
@@ -40,15 +41,38 @@ export class DashboardStateEngine {
...
+        let isFullTimeSelection =
```
neogucky, line 56:
```suggestion
        const isFullTimeSelection =
```

projects/step-frontend/src/lib/modules/timeseries/components/dashboard/dashboard-state-engine.ts
```diff
@@ -40,15 +41,38 @@ export class DashboardStateEngine {
...
+        timeRangeSettings.fullRange = fullRange;
+        if (isFullTimeSelection) {
+          timeRangeSettings.selectedRange = { ...timeRangeSettings.fullRange };
+        } else {
+          // we have a custom selection. has to be cropped if needed
+          const currentSelection = timeRangeSettings.selectedRange;
+          const newFrom = Math.max(fullRange.from!, currentSelection.from!);
+          const newTo = Math.min(fullRange.to!, currentSelection.to!);
+          if (newTo - newFrom < 3) {
+            timeRangeSettings.selectedRange = { ...timeRangeSettings.fullRange }; // zoom reset when the interval is very small
```
neogucky, line 69:
```suggestion
            // zoom reset when the interval is very small
            timeRangeSettings.selectedRange = { ...timeRangeSettings.fullRange }; 
```

neogucky: The 3 looks quite random to me, why must the zoom be reset for 2 but not for lets say 5?

projects/step-frontend/src/lib/modules/timeseries/components/dashboard/dashboard-state-engine.ts
```diff
@@ -82,11 +106,44 @@ export class DashboardStateEngine {
...
+        newSettings = {
+          type: TimeRangeType.RELATIVE,
+          defaultFullRange: oldSettings.defaultFullRange,
+          fullRange: { from: from, to: now },
+          selectedRange: { from: from, to: now },
+          relativeSelection: relativeSelection,
+        };
+        break;
+    }
     const state = this.state;
```
neogucky, line 145: Do we need this?

neogucky (changes requested):
Check my comments. Also it seems like your switch/cases don't cover the dafault case. Is it technically impossible that the state is other than the three defined (as the value is an enum with only these values) or is there a chance it is not one of them? In that case I would cover it.

projects/step-frontend/src/lib/modules/timeseries/components/dashboard/dashboard-state-engine.ts
lucianbaciu: @neogucky I added a variable and also added a comment. I tried to explain but I don't remember exactly the context, but I've used that for a reason

lucianbaciu:
@neogucky the switches on time range always cover all 3 options. I definitely prefer to use default when at least once options is not covered

neogucky (approved):

PR #910, opened 2024-08-07
SED-3222 add custom toast

lukasz-lepa:
https://exense.atlassian.net/browse/SED-3304

### DIFF
--- a/projects/step-core/src/lib/components/toast-container/toast-container.component.ts
+++ b/projects/step-core/src/lib/components/toast-container/toast-container.component.ts
@@ -1,0 +1,20 @@
+import { Component, OnInit } from '@angular/core';
+import { ToastService } from '../../services/toast.service';
+import { NotificationAction } from '../../shared/toast-action.interface';
+
+@Component({
+  selector: 'step-toast-container',
+  templateUrl: './toast-container.component.html',
+  styleUrls: ['./toast-container.component.scss'],
+})
+export class ToastContainerComponent implements OnInit {
+  toasts: { message: string; actions?: NotificationAction[]; duration?: number }[] = [];
+
+  constructor(private toastService: ToastService) {}
+
+  ngOnInit(): void {
+    this.toastService.toastMessages.subscribe((toasts) => {
+      this.toasts = toasts;
+    });
+  }
+}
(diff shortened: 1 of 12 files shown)


projects/step-core/src/lib/components/toast-container/toast-container.component.ts
```diff
@@ -0,0 +1,20 @@
...
+export class ToastContainerComponent implements OnInit {
+  toasts: { message: string; actions?: NotificationAction[]; duration?: number }[] = [];
+
+  constructor(private toastService: ToastService) {}
+
+  ngOnInit(): void {
+    this.toastService.toastMessages.subscribe((toasts) => {
+      this.toasts = toasts;
+    });
+  }
```
dvladir, line 19: Please, check the comment below in ToastService. With signal usage the code of current component could look like this
```suggestion
protected readonly _toasts = inject(ToastService).toastMessages;
```

projects/step-core/src/lib/components/toast/toast.component.scss
```diff
@@ -0,0 +1,36 @@
+.toast {
```
dvladir, line 1: Please use `rem` instead of `px`
Use colors variables defined in `core-variables`

projects/step-core/src/lib/components/toast/toast.component.html
```diff
@@ -0,0 +1,14 @@
+@if (message) {
+  <div class="toast">
+    <span class="success-icon">&#10003;</span>
```
dvladir, line 3: @neogucky @lukasz-lepa 
Shouldn't we use feather-icons here instead of unicode symbols?

projects/step-core/src/lib/components/toast/toast.component.html
```diff
@@ -0,0 +1,14 @@
...
+    <p class="toast-message">{{ message }}</p>
+    <div class="toast-actions">
+      @for (action of actions; track action) {
```
dvladir, line 6:
```suggestion
      @for (action of actions; track action.label) {
```

dvladir (changes requested):

PR #922, opened 2024-08-20
SED-3326 fixed browser back broken for executions

AmirBenkoula:
Going back did not work because execution button leads to `/executions/open/${id}` which replaces the history.


neogucky (changes requested):
Thanks, this solution works indeed. However I don't like to set the "steps" in this place for two reasons:

1. we also handle the tab initialization here: private ExecutionProgressComponent.initTabs(): void and would thus have a logic doublication
2. we are currently planning a new execution details page (you didn't know about this yet) and will have a different structure, there the direct navigation to "steps" wouldn't work

I took a look at the code and would suggest the following changes:
- keep the replaceUrl: true from your solution
- then in initTabs, add another parameter to the selectTab call to let it knwo this is an init call
- add a new signitaure ExecutionProgressComponent.selectTab(tabId: string, **isInitialize: boolean**): void
and also use replaceUrl: true in the case isInitialize is true

neogucky (approved):

PR #923, opened 2024-08-22
SED-3339 create cross execution dashboard page

AmirBenkoula:
- Page in /#/executions/schedules/SCHEDULE_ID/ with icon and schedule name
- Link to the page in schedule menu


projects/step-frontend/src/lib/modules/execution/components/schedule-overview/schedule-overview.component.html
```diff
@@ -0,0 +1,4 @@
+<div>
+  <step-icon name="clock"></step-icon>
```
neogucky, line 2: This is totally correct for the current status, but as we might change the scheduler icon at some point we need to do it a bit more dynamically. Check out <entity-icon> (step-frontend/projects/step-core/src/lib/modules/entity/components/entity-icon/entity-icon.component.ts) 
We need to provide the entityName and also the entity object

projects/step-frontend/src/lib/modules/execution/components/schedule-overview/schedule-overview.component.ts
```diff
@@ -0,0 +1,36 @@
...
+import { ActivatedRoute } from '@angular/router';
+import { AugmentedSchedulerService, ExecutiontTaskParameters } from '@exense/step-core';
+
+@Component({
+  selector: 'step-schedule-overview',
+  templateUrl: './schedule-overview.component.html',
+  styleUrls: ['./schedule-overview.component.scss'],
+})
+export class ScheduleOverviewComponent implements OnInit {
+  private _api = inject(AugmentedSchedulerService);
```
neogucky, line 11: As we will later also have other API services in this controller (at least AugmentedExecutionSerivice) we should already start naming the services with good names like _scheduleApi

projects/step-frontend/src/lib/modules/execution/components/schedule-overview/schedule-overview.component.ts
```diff
@@ -0,0 +1,36 @@
...
+        this._api.getExecutionTaskById(this.taskId).subscribe({
+          next: (task) => (this.task = task),
+          error: () => {
+            this.error = 'Invalid Task Id or server error.';
+          },
+        });
+      } else {
+        this.error = 'Task Id not found in the URL.';
+      }
+    });
```
neogucky, line 34: This code works, but we have a slightly more complicated but more robust method of extracting entities from the url, check how we do it for the edit task dialog:

step-frontend/projects/step-core/src/lib/modules/scheduler-common/types/edit-scheduled-task.route.ts

neogucky (changes requested):
Good work for the second ticket, but check my comments for improvements!

projects/step-frontend/src/lib/modules/execution/components/schedule-overview/schedule-overview.component.ts
```diff
@@ -0,0 +1,75 @@
...
+            this.initializeTask();
+          },
+          error: () => {
+            this.error = 'Invalid Task Id or server error.';
+          },
+        });
+      } else {
+        this.error = 'Task Id not found in the URL.';
+      }
+    });
```
dvladir, line 42: Please avoid `subscribe` inside `subscribe`.
Refactor this observable chain with usage of `switchMap` operator.
Also `route.url` is hot observable - it's need to be unsubscribed, when component is destroyed. It's can be done with `takeUntilDestroyed` operator

projects/step-frontend/src/lib/modules/execution/components/schedule-overview/schedule-overview.component.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  protected task = this._taskData;
+
+  protected plan?: Partial<Plan>;
+
+  protected error = '';
+  protected repositoryId?: string;
+  protected repositoryPlanId?: string;
+
+  constructor(private route: ActivatedRoute) {}
```
dvladir, line 23: `inject` syntax is preferred.
Mark injected fields with `_` prefix

projects/step-frontend/src/lib/modules/execution/components/schedule-overview/schedule-overview.component.ts
```diff
@@ -0,0 +1,75 @@
...
+        });
+      } else {
+        this.error = 'Task Id not found in the URL.';
+      }
+    });
+  }
+
+  private initializeTask(): void {
+    if (this.task) {
+      if (!this.task.attributes) {
```
dvladir, line 47: Block's depth can be reduced.
As method's body make sense when task exists only - it can be written as:
```
private initializeTask(): void {
   if (!this.task) {
       return;
   }
   
   // rest of the code
  .... 
}
```

projects/step-frontend/src/lib/modules/execution/components/schedule-overview/schedule-overview.component.ts
```diff
@@ -0,0 +1,75 @@
...
+
+@Component({
+  selector: 'step-schedule-overview',
+  templateUrl: './schedule-overview.component.html',
+  styleUrls: ['./schedule-overview.component.scss'],
+})
+export class ScheduleOverviewComponent implements OnInit {
+  private _scheduleApi = inject(AugmentedSchedulerService);
+  private taskId?: string;
+  private _taskData?: ExecutiontTaskParameters;
```
dvladir, line 13: Underscore prefix is related to injections only

dvladir (dismissed):

PR #947, opened 2024-09-25
SED-3284 extended info-banner and added initial banners

neogucky:
Some links are still missign and texts could be updated before merging!


projects/step-core/src/lib/modules/info-banner/injectables/info-banner-register.service.ts
```diff
@@ -28,6 +28,7 @@ export class InfoBannerRegisterService {
   }
 
   private registerRouteResolver(route: Route, infoBannerKey: string): void {
+    console.log('INFOBANNER', infoBannerKey, route);
```
neogucky, line 31:
```suggestion
```

PR #949, opened 2024-09-27
SED-3391 plan editor does not allow to set dynamic names for call keyword and call plan components

AmirBenkoula:


projects/step-core/src/lib/components/reference-artefact-name/reference-artefact-name.component.html
```diff
@@ -1,41 +1,62 @@
...
-    </button>
-  }
+  <button
+    type="button"
+    [matTooltip]="_artefactNameConfig.captions.editSelectionCriteria"
+    matTooltipPosition="below"
+    mat-icon-button
+    [color]="isEditorMode ? 'primary' : undefined"
+    (click)="isEditorMode = !isEditorMode"
+    [ngClass]="{ 'dynamic-on': artefact?.dynamicName?.dynamic === true }"
```
dvladir, line 36: Just a hint. For simple cases, when it's required to conditionally add only one class the following syntax looks more compact:
`[class.dynamic-on]="!!artefact?.dynamicName?.dynamic"`

PR #954, opened 2024-10-01
SED-3326 Dialog close behaivor on history back

dvladir:
The main issue occurred, because `exitRoute` method was invoked before route's navigation events were performed in case of history back actions.
It caused the situation, that navigation in `exitRoute` overrode browser history with new items.
Applying `asyncScheduler` fixed it.
In case of back navigation - route event's performs first. It destroys the component and unsubscribe from dialog close events. `exitRoute` no longer invoked in this case and doesn't spoil the history, with unnecessary navigation.


AmirBenkoula (approved):
Everything looks good to me :)

neogucky (approved):

PR #972, opened 2024-10-25
SED-3383 added slide toggle component when checkbox is selected

AmirBenkoula:


projects/step-frontend/src/lib/modules/admin/components/screen-input-edit-dialog/screen-input-edit-dialog.component.html
```diff
@@ -34,7 +34,8 @@
         <step-label-addon>
           <step-icon name="help-circle" matTooltip="Defines the default value of this input field (Optional)" />
         </step-label-addon>
-        <input type="text" [(ngModel)]="model.defaultValue" />
+        <input *ngIf="model.type !== 'CHECKBOX'" type="text" [(ngModel)]="model.defaultValue" />
+        <mat-slide-toggle color="primary" *ngIf="model.type === 'CHECKBOX'" [(ngModel)]="model.defaultValue" />
```
neogucky, line 38: @AmirBenkoula please use the new @if syntax and also use @else for the second case. The reason this is important is so it is clear there are only these two cases possible (currently a dev needs to understand the content of the if clause to understand this)

neogucky (changes requested):
see comment

neogucky (approved):

PR #985, opened 2024-11-08
SED-3554-create pie chart standalone component

lucianbaciu:


neogucky (approved):

projects/step-frontend/src/lib/modules/timeseries/components/doughnut-chart/doughnut-chart.component.ts
```diff
@@ -0,0 +1,81 @@
...
+  templateUrl: './doughnut-chart.component.html',
+  styleUrls: ['./doughnut-chart.component.scss'],
+  imports: [StepCommonModule],
+  standalone: true,
+})
+export class DoughnutChartComponent implements AfterViewInit, OnDestroy, OnChanges {
+  @Input() settings!: DoughnutChartSettings;
+
+  @ViewChild('canvas') private canvas!: ElementRef<HTMLCanvasElement>;
+  private chart: Chart | undefined;
```
dvladir, line 28: @lucianbaciu Please use signals API for inputs and viewChild.

chart itself, will be computed signal in this case

dvladir (changes requested):

PR #1001, opened 2024-12-11
SED-3623-add input for relative time ranges

lucianbaciu:


lucianbaciu:
I tried to use signals input but they can't be updated. DIdn't find any option to pass the inputs otherwise than manually selecting them

neogucky (approved):

PR #1018, opened 2025-01-22
SED-3704-added nice loading

lucianbaciu:


neogucky (commented):
Please use '' when marking an empty json-string (instead of `` which is used in some places)

Also, why where the titles removed? Not saying it's wrong, I just want to understand it

lucianbaciu:
@neogucky The titles were removed because I use the title of the widget instead (so all piecharts, charts and tables have the same header look)

neogucky (approved):

PR #1043, opened 2025-02-26
SED-3766-fix issue

lucianbaciu:


projects/step-frontend/src/lib/modules/execution/components/alt-execution-progress/alt-execution-progress.component.ts
```diff
@@ -458,8 +458,14 @@ export class AltExecutionProgressComponent implements OnInit, OnDestroy, AltExec
...
         let time = selection.relativeSelection!.timeInMs;
-        let end = execution.endTime || new Date().getTime() - 5000;
-        selection!.absoluteSelection = { from: end - time, to: end };
+        let now = new Date().getTime();
+        let end = execution.endTime || now - 5000;
+        let from = end - time;
+        if (from > end) {
+          // remove the 5 sec buffer
+          end = now;
+        }
```
neogucky, line 467: I don't really understand this buffer, why do we need it in the first place? And why is it save to remove in case its smaller than 5 seconds?

neogucky (approved):

PR #1088, opened 2025-04-03
SED-3773 Add partial execution tree to node details view

dvladir:


neogucky (dismissed):

neogucky (approved):

projects/step-frontend/src/lib/modules/execution/components/alt-keyword-inline-drilldown/alt-report-node-details.component.html
```diff
@@ -1,45 +1,78 @@
...
-  }
+  </div>
+  <div class="tree-container execution-report-node-details">
+    <div class="data-container">
+      <div class="container-title">
+        <step-icon name="tree" />
+        <span>Partial execution tree</span>
+        <step-icon
+          name="alert-circle"
+          matTooltip="The partial execution tree of this node shows the branch leaging to it's direct siblings"
```
neogucky, line 54:
```suggestion
          matTooltip="This is a partial execution tree showing the selected node in context. It includes the path from the root to the selected node, the selected node’s children, and — if the node is part of an iterative structure (like a ThreadGroup) — its direct siblings to provide full loop context."
```

neogucky (approved):

PR #1112, opened 2025-04-23
SED-3942 Improve current attachement implementation

dvladir:


projects/step-core/src/lib/modules/attachments/components/alt-attachment-preview/alt-attachment-preview.component.scss
```diff
@@ -0,0 +1,101 @@
...
+step-alt-attachment-preview {
+  display: block;
+  width: 15rem;
+  height: 7rem;
+  position: relative;
+  cursor: pointer;
+  overflow: hidden;
+
+  &.large {
+    width: 60rem;
```
neogucky, line 13:
```suggestion
    width: 45rem;
```

projects/step-core/src/lib/modules/attachments/components/alt-attachment-preview/alt-attachment-preview.component.scss
```diff
@@ -0,0 +1,101 @@
...
+    height: 28rem;
```
neogucky, line 14:
```suggestion
    height: 21rem;
```

neogucky (changes requested):
@dvladir two archtectual naming things:

- As we want to refactor the naming of the execution report view to remove "alt" prefixes we shouldn't start adding new ones. If there is a conflict with the legacy report view please try to rename the legacy component with "legacy" prefix instead and use the clean name for the current version.

- We now have both "alt-attachement-preview" and "alt-attachement-previews" which are dificult to differentiate. Instead of plural we should use "-list"

dvladir:
> @dvladir two archtectual naming things:
> 
>     * As we want to refactor the naming of the execution report view to remove "alt" prefixes we shouldn't start adding new ones. If there is a conflict with the legacy report view please try to rename the legacy component with "legacy" prefix instead and use the clean name for the current version.
> 
>     * We now have both "alt-attachement-preview" and "alt-attachement-previews" which are dificult to differentiate. Instead of plural we should use "-list"

@neogucky Components have been renamed

neogucky (approved):

PR #1114, opened 2025-04-24
SED-3926 Improve rendering of execution report nodes

dvladir:


projects/step-core/src/lib/modules/artefacts-common/components/artefact-inline-field/artefact-inline-field.component.ts
```diff
@@ -1,8 +1,9 @@
 import { ChangeDetectionStrategy, Component, computed, inject, input, ViewEncapsulation } from '@angular/core';
-import { NgClass } from '@angular/common';
 import { AceMode, RichEditorDialogService } from '../../../rich-editor';
 import { ArtefactInlineItem } from '../../types/artefact-inline-item';
-import { StepBasicsModule } from '../../../basics/step-basics.module';
+import { PopoverMode, StepBasicsModule } from '../../../basics/step-basics.module';
+
+const DEFAULT_MARGIN = '0.5rem';
```
neogucky, line 6: I'm generally not a fan of styles being defined in the controller, but here this is probably the best way. Though I would like to clearly mark such info (as well as the margin signal) so it is obviously apparent. I will add a discussion point to Trello.

neogucky (approved):

neogucky (approved):

PR #1148, opened 2025-06-17
SED-4002 Add dynamic sizing for execution details

dvladir:


projects/step-core/src/lib/modules/artefacts-common/components/artefact-inline-field-list/artefact-inline-field-list.component.ts
```diff
@@ -13,52 +32,177 @@ import { StepBasicsModule } from '../../../basics/step-basics.module';
...
+    const displayItems = this.displayItems();
+    return displayItems.length < items.length;
   });
+
+  private determineElementsWithWidths(
+    renderedElementsWithInitialWidths: ArtefactInlineFieldComponent[],
+    availableWidth: number,
+    isLog?: boolean,
+  ): ArtefactInlineItemExplicitWidths[] {
+    let wholeTotal = this.countTotal(renderedElementsWithInitialWidths.map((element) => element.getWidths()));
```
neogucky, line 85: I don't like the naming, wholeTotal should be something like "totalWidth"

projects/step-core/src/lib/modules/artefacts-common/components/artefact-inline-field-list/artefact-inline-field-list.component.ts
```diff
@@ -13,52 +32,177 @@ import { StepBasicsModule } from '../../../basics/step-basics.module';
...
+
+    let changedItems: ArtefactInlineItemExplicitWidths[] = [];
+    let areElementsCountDecreased = false;
+
+    while (wholeTotal > availableWidth && renderedElementsWithInitialWidths.length > 0) {
+      const totalCount = renderedElementsWithInitialWidths.reduce((res, item) => res + (item.getCount() ?? 0), 0);
+      let fareShare = Math.round(availableWidth / totalCount);
```
neogucky, line 92: I think you used a typo from David, it should probably be "fairShare"

PR #1161, opened 2025-07-03
Sed 3939 larger input field for webhook body

neogucky:


neogucky:
These are old changes - it seems they where not needed for the merge. deleting for now.

PR #1192, opened 2025-08-04
SED-4111 Update the detail panel when clicking on another node in execution tree

dvladir:


projects/step-frontend/src/lib/modules/execution/components/aggregated-tree-node-dialog/aggregated-tree-node-dialog.component.ts
```diff
@@ -32,7 +33,13 @@ export interface AggregatedTreeNodeDialogData {
   templateUrl: './aggregated-tree-node-dialog.component.html',
   styleUrl: './aggregated-tree-node-dialog.component.scss',
   encapsulation: ViewEncapsulation.None,
-  providers: [AltExecutionDialogsService],
+  providers: [
+    {
+      provide: NODE_DETAILS_RELATIVE_PARENT,
+      useFactory: () => inject(ActivatedRoute).parent!.parent!,
```
neogucky, line 39: @dvladir are we sure that we always have .parent!.parent!?

neogucky (approved):

projects/step-frontend/src/lib/modules/execution/components/aggregated-tree-node-dialog/aggregated-tree-node-dialog.component.ts
neogucky: No, this is fine!

PR #1194, opened 2025-08-06
SED-4009-added placeholder for deleted plans/tasks

lucianbaciu:


projects/step-frontend/src/lib/modules/execution/components/schedule-overview/cross-execution-dashboard/cross-execution-dashboard.component.html
```diff
@@ -1,28 +1,16 @@
 <section class="header" [class.small-screen]="!!(_isSmallScreen$ | async)">
   <div class="title">
-    @if (_state.task(); as task) {
-      <h3 data-step-testid>
+    <h3 data-step-testid>
```
neogucky, line 3: I know you didn't add this now, but lets fix empty data-step-testids when touching files. I would use "task-or-plan-title"

projects/step-frontend/src/lib/modules/execution/components/schedule-overview/cross-execution-dashboard/cross-execution-dashboard.component.html
```diff
@@ -1,28 +1,16 @@
...
-          >
-            <step-icon name="external-link" />
-          </a>
-        }
-      </h3>
-    }
-    @if (_state.plan(); as plan) {
-      <h3 data-step-testid>
+      }
+      @if (_state.plan(); as plan) {
```
neogucky, line 7:
```suggestion
      @else if (_state.plan(); as plan) {
```

projects/step-frontend/src/lib/modules/execution/components/schedule-overview/cross-execution-dashboard/cross-execution-dashboard.component.ts
```diff
@@ -24,6 +24,29 @@ export class CrossExecutionDashboardComponent implements OnInit {
...
+      const plan = this._state.plan();
+      if (plan === undefined) {
+        return 'Loading...';
+      } else if (plan === null) {
+        return 'Plan name (deleted)';
+      } else {
+        return plan.attributes?.['name'];
+      }
+    }
+  });
```
neogucky, line 48: Not sure I understand why we are adding "name" here. This is shown in the top of the view, correct? In that case I would write "Deleted task" / "Deleted plan"

Also the code could be improved as all except for the deleted part is the same (please check as I wrote it in the comment field untested):

> const isTask = this._state.viewType() === 'task';
> let entity = isTask ? this._state.task() : this._state.plan();
>       if (entity === undefined) {
>         return 'Loading...';
>       } else if (entity === null) {
>         return isTask ? "Deleted task" : "Deleted plan";
>       } else {
>         return entity?.attributes?.['name'];
>       }
>

projects/step-frontend/src/lib/modules/execution/components/schedule-overview/scheduler-page/scheduler-page.component.ts
```diff
@@ -50,7 +50,13 @@ export class SchedulerPageComponent implements OnInit {
 
   ngOnInit(): void {
     this._schedulerService.getExecutionTaskById(this._taskIdFn()).subscribe((task) => {
-      this.state.task.set(task);
+      console.log('task loaded', task);
```
neogucky, line 53:
```suggestion
```

projects/step-frontend/src/lib/modules/execution/components/schedule-overview/scheduler-page/scheduler-page.component.ts
```diff
@@ -50,7 +50,13 @@ export class SchedulerPageComponent implements OnInit {
...
+      if (task === undefined) {
+        // the task was not found
+        this.state.task.set(null);
+      } else {
+        this.state.task.set(task);
+      }
```
neogucky, line 59: I noticed that you introduced "null" as an additional value for the task and plan. Why don't we just use undefined? I assume you tried that but I would like to understand ;)

neogucky (changes requested):

neogucky (approved):

PR #1207, opened 2025-08-20
SED-4176 Select all issue with test case selection

dvladir:


projects/step-core/src/lib/modules/entities-selection/components/bulk-sel/bulk-sel.component.html
neogucky, line 1: same as for the label: "bulk-selection"

PR #1228, opened 2025-09-12
SED-4218 Improve debugging of errors in execution tree

dvladir:


projects/step-core/src/lib/modules/search/components/search-field/search-field.component.ts
```diff
@@ -51,6 +68,9 @@ export class SearchFieldComponent implements ControlValueAccessor {
     this.searchIndex.set(0);
   });
 
+  private fieldAddon = contentChild(SearchFieldAddonDirective);
+  protected readonly tplFieldAddon = computed(() => this.fieldAddon()?._templateRef);
```
neogucky, line 72: I don't like the name "tplFieldAddon", for me it is not immediately clear what this stands for

neogucky (dismissed):

neogucky (approved):

PR #1341, opened 2026-03-12
SED-4543 Drilldown workflow in execution tree

dvladir:


neogucky (approved):
Approved removal of tree tab - later we will have to do a new review of follow-up tasks related to drilldown

PR #1350, opened 2026-04-15
SED-3986 Improve icons for controls

neogucky:


custom-icons/assert-metric.svg
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg xmlns="http://www.w3.org/2000/svg" version="1.1" viewBox="0 0 24 24"  class="feather feather-assert-metric">
+  <defs>
+    <style>
+      .st0 {
```
dvladir, line 5: @neogucky Such style classes names need to be renamed, because they are included globally and might affect on each-other.

Check how they are declared for some of the existing icons:
managed-file.svg or binary-file.svg

dvladir (changes requested):

dvladir (approved):

PR #1359, opened 2026-05-01
SED-4166 cross executions report for all plan repositories

lucianbaciu:


.husky/pre-commit
dvladir, line 1: We shouldn't remove this file!
Restore it back.

projects/step-core/src/lib/modules/execution-common/pipes/execution-name.pipe.ts
```diff
@@ -1,5 +1,5 @@
 import { Pipe, PipeTransform } from '@angular/core';
-import { Execution } from '../../../client/step-client-module';
+import { Execution } from '../../../client/generated';
```
dvladir: Import from module is preferred.
```suggestion
import { Execution } from '../../../client/step-client-module';
```

projects/step-frontend/src/lib/modules/execution/components/execution-list/execution-list.component.html
```diff
@@ -170,9 +170,51 @@
...
+          </button>
+          <a
+            stepTestId="cross-execution-plan-btn"
+            mat-icon-button
+            matTooltip="Open plan report"
+            [disabled]="!('plan-read' | hasRight | async)"
+            [routerLink]="element.executionParameters?.repositoryObject && element.executionParameters?.repositoryObject?.repositoryID !== 'local' ? (element.id | crossExecutionRepositoryUrl) : (element.planId | crossExecutionPlanUrl)"
+          >
+            <step-icon name="file-07" />
+          </a>
```
dvladir, line 213: To simplify the condition in routerLink `let` syntax can be used:

```suggestion
         @let repositoryObject = element.executionParameters?.repositoryObject;
         @let isNotLocal = !!repositoryObject && repositoryObject.repositoryID !== 'local';
          <a
            stepTestId="cross-execution-plan-btn"
            mat-icon-button
            matTooltip="Open plan report"
            [disabled]="!('plan-read' | hasRight | async)"
            [routerLink]="isNotLocal  ? (element.id | crossExecutionRepositoryUrl) : (element.planId | crossExecutionPlanUrl)"
          >
            <step-icon name="file-07" />
          </a>
```

projects/step-frontend/src/lib/modules/execution/components/schedule-overview/cross-execution-dashboard/report/scheduler-report-view.component.ts
dvladir, line 1: If members of this component are invoked from it's template only, make sense to mark them as protected

dvladir (changes requested):

projects/step-core/src/lib/client/generated/services/ExecutionsService.ts
```diff
@@ -458,4 +458,15 @@ export class ExecutionsService {
...
+  public getLastExecutionsByPlan(planId: string, limit: number, from?: number, to?: number): Observable<Execution[]> {
+    return this.httpRequest.request({
+      method: 'GET',
+      url: `/executions/search/last/by/plan/${planId}`,
+      mediaType: 'application/json',
+      query: {
+        limit, from, to
+      }
+    });
+  }
```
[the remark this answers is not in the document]
dvladir: @lucianbaciu Was this method autogenerated or manually added?
I don't see method's descritpions comments that are usually places by openapi-codegen.

If it was added manually, we should avoid it, because next automatic generation will erase it. Instead mtehod should be defied in OpenAPI schema.

If it was generated from schema, probably something is missing in annotation descriptions, because there are no comments.

Please double check it.

projects/step-frontend/src/lib/modules/execution/components/schedule-overview/cross-execution-dashboard/cross-execution-dashboard-state.ts
```diff
@@ -489,13 +491,23 @@ export abstract class CrossExecutionDashboardState {
...
 
   errorTableRefreshSub = this.timeRange$.subscribe((timeRange) => {
-    let filterItem = this.getDashboardFilter();
-    // this is working only with searchEntities for now. extend it if needed
-    const filter = { [filterItem.attributeName]: filterItem.searchEntities[0]?.searchValue };
-    this.errorsDataSource.reload({
-      request: { timeRange: timeRange, ...filter },
-      hideProgress: this.lastRefreshTrigger() === 'auto',
-    });
+    let entityParams = undefined;
```
dvladir, line 494: `timeRange$` is observable, `this.execution()` is signal
Should the logic of this subscription react when both are changing or, when timeRange is changing only?

1. In case of both change tracking, this subscription should be refactored. Convert `timeRange` to signal, or `execution` to observable. Track changes of both of them with signal's effect or rxjs combine latest
2. In it needs to track the `timeRange` only, get execution and canonicalPlan name in the following way:
```
const execution = untracked(() => this.execution());
const canonicalPlanName = execution?.importResult?.canonicalPlanName;
...
switch(this.getViewType()) {
   ...
   case 'repository':
      if (!canonicalPlanName) {
          return;
      } 
      entityParams = { canonicalPlanName };
   break;
} 
```

lucianbaciu: it only needs to trigger when timeRange$ changes. I don't see the reason to use untracked since I'm not in an effect/compute method

dvladir: @lucianbaciu Ok!

generate-openapi-client.js
```diff
@@ -88,8 +88,8 @@ async function main() {
 
   shell.sed('-i', '.*Service.*', '', `${GENERATED_CODE_LOCATION}/${CLIENT_MODULE_NAME}.ts`);
   shell.sed('-i', '/rest', 'rest', `${GENERATED_CODE_LOCATION}/StepGeneratedClientModule.ts`);
-  shell.exec(`npx prettier --write ${GENERATED_CODE_LOCATION}/*`);
-  shell.exec(`npx prettier --write ${GENERATED_CODE_LOCATION}/**/*`);
+  shell.exec(`npx prettier --end-of-line crlf --write ${GENERATED_CODE_LOCATION}/*`);
+  shell.exec(`npx prettier --end-of-line crlf --write ${GENERATED_CODE_LOCATION}/**/*`);
```
dvladir: @lucianbaciu This is not required.
In git remote repository the `lf` ending is used. On Linux/Max machines also.
In case of Windows the git parameter `core.autocrlf`should be set to true - to automatically convert endings to appropriate format during check out and commit

dvladir (dismissed):

PR #1370, opened 2026-05-22
SED-4676 Global project entities cannot be used from other projects

dvladir:


neogucky (changes requested):
Please remove the console.log (see geminis comment) afterwards can be approved

neogucky (approved):

PR #1378, opened 2026-05-28
SED-4714-step-30-analytics-finalization-before-release

david-stephan:


neogucky (approved):
Looks good! The gemini comments about the icon are not important in my opinion, but tracking the dashlets by id instead of name would be a good practise.

neogucky (dismissed):
Looks good! The gemini comments about the icon are not important in my opinion, but tracking the dashlets by id instead of name would be a good practise.

PR #1379, opened 2026-05-29
SED-4743 restored attachment sizing

neogucky:


projects/step-core/src/lib/modules/attachments/components/attachment-inline-preview-list/attachment-inline-preview-list.component.ts
```diff
@@ -117,6 +124,10 @@ export class AttachmentInlinePreviewListComponent implements AfterViewInit {
     return result;
   }
 
+  private applyInitialWidths(items: AttachmentMetaWithExplicitWidth[]): AttachmentMetaWithExplicitWidth[] {
```
lucianbaciu, line 127: I'd not necessarly add this method if it is used only one place

lucianbaciu (approved):

projects/step-core/src/lib/modules/attachments/components/attachment-inline-preview-list/attachment-inline-preview-list.component.ts
neogucky: The intent was to improve code readability, but generally I agree. Though as it is minor I don't want to push the merge back further so lets leave it for now.

PR #1387, opened 2026-06-02
SED-4733 details node history does not support canonical plan name

lucianbaciu:


projects/step-frontend/src/lib/modules/execution/components/aggregated-tree-node-history/aggregated-tree-node-history.component.ts
```diff
@@ -46,12 +46,31 @@ export class AggregatedTreeNodeHistoryComponent {
...
-    ),
+    switchMap((ex: Execution) => {
+      let lastExecutions: Observable<Execution[]>;
+      if (ex.executionTaskID) {
+        lastExecutions = this.fetchLastExecutionsByTask(ex.startTime! - 1, ex.executionTaskID);
+      } else {
+        const repositoryObject = ex.executionParameters?.repositoryObject;
+        const isNotLocal = !!repositoryObject && repositoryObject.repositoryID !== 'local';
+        if (isNotLocal) {
+          let canonicalPlanName = ex.importResult?.canonicalPlanName;
```
david-stephan: @lucianbaciu I do not understand this section, since we anyway do not show the node history for external repo, we should not have to check if there is a canonical plan name and should not need to fetch the execution for them.

neogucky (changes requested):
Please check the gemini comments, as they sound correct to me (I didn't go through the full pipe to check it though)

david-stephan (changes requested):
See my comment, for the gemini comments, I think that's what we wanted to do, so I guess they can be ignored.

david-stephan (approved):

PR #1399, opened 2026-06-18
SED-4773 The Project column disappeared in the Execution view

neogucky:


projects/step-core/src/lib/modules/table/services/table-columns.service.ts
```diff
@@ -292,21 +292,41 @@ export class TableColumnsService implements Reloadable, OnDestroy {
...
+    columnSettingList.forEach((col, i) => (col.position = i));
+    columnsToAdd
+      .sort((a, b) => (a.position ?? 0) - (b.position ?? 0))
+      .forEach((colToAdd) => {
+        const insertPosition = Math.min(colToAdd.position ?? columnSettingList.length, columnSettingList.length);
+        columnSettingList.forEach((col) => {
+          if ((col.position ?? 0) >= insertPosition) {
+            col.position = (col.position ?? 0) + 1;
+          }
+        });
```
dvladir, line 324: `col.position ?? 0` looks redundant as it's position has been explicitly set  on row 314
```suggestion
        columnSettingList.forEach((col) => {
              if (col.position >= insertPosition) col.position++;
        });
```

dvladir (changes requested):

PR #1404, opened 2026-06-24
SED-4785 Screentemplates with empty attributes break view

neogucky:


projects/step-frontend/src/lib/modules/admin/components/screen-input-edit-dialog/screen-input-edit-dialog.component.ts
```diff
@@ -24,16 +25,27 @@ export class ScreenInputEditDialogComponent {
...
-  readonly activationExpressionFormatter = EXPRESSION_SCRIPT_FORMATTER;
+  protected readonly customUIComponentsFormatter = CUSTOM_UI_COMPONENTS_FORMATTER;
+  protected readonly activationExpressionFormatter = EXPRESSION_SCRIPT_FORMATTER;
+  protected readonly errorsDictionary: Record<string, string> = {
+    required: 'This field is required',
+  };
+
+  private readonly attributeInput = viewChild<NgModel>('attributeInput');
+
+  @HostListener('keydown.enter')
```
dvladir: @neogucky Angular documentation recommends to use host properties instead of HostListerner and HostBinding.
Check the warning at the bottom of the section https://angular.dev/guide/components/host-elements#the-hostbinding-and-hostlistener-decorators

I would prefer to avoid HostListerner in the new code

dvladir (changes requested):

PR #1409, opened 2026-07-03
SED-4782 Attachment viewer for .md (markdown) files

neogucky:


projects/step-core/src/lib/modules/attachments/components/attachment-dialog/attachment-dialog.component.scss
```diff
@@ -153,6 +153,76 @@ step-attachment-dialog {
...
+    background: var.$white;
+    color: var.$gray-700;
+
+    &.loading markdown {
+      visibility: hidden;
+    }
+
+    markdown {
+      display: block;
+      max-width: 960px;
```
dvladir: Should be 96rem

projects/step-core/src/lib/modules/attachments/components/attachment-dialog/attachment-dialog.component.scss
```diff
@@ -153,6 +153,76 @@ step-attachment-dialog {
...
+    pre,
+    code {
+      font-family: var.$code-font;
+    }
+
+    pre {
+      overflow: auto;
+      padding: 1rem;
+      background: var.$gray-100;
+      border-radius: 4px;
```
dvladir: We agreed to always use rems. This style changes need to be double checked and improved.
1rem = 10px in our case. So instead of 4px should be 0.4rem

package.json
```diff
@@ -64,6 +64,8 @@
     "angular-split": "19.0.0",
     "feather-icons": "4.29.0",
     "luxon": "3.4.4",
+    "marked": "15.0.12",
+    "ngx-markdown": "19.1.1",
```
dvladir, line 68: These two dependencies should be also added to EE package.json
Plus it should be defined in `shared-libraries.config.js` files both in OS and EE

dvladir (changes requested):

projects/step-core/src/lib/modules/attachments/components/attachment-dialog/attachment-dialog.component.scss
```diff
@@ -153,6 +153,76 @@ step-attachment-dialog {
...
+      border-collapse: collapse;
+      width: 100%;
+    }
+
+    th,
+    td {
+      padding: 0.5rem;
+      border: 1px solid var.$gray-300;
+    }
+  }
```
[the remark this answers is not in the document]
dvladir: This component has ViewEncapsulation.NONE

PR #1413, opened 2026-07-06
SED-4763 Bar charts - improve the bar edges

neogucky:


lucianbaciu (dismissed):
the code looks good. I also tested with different situation and it works good!

PR #1420, opened 2026-07-13
SED-4005-added next execution date in scheduler page and list

lucianbaciu:


projects/step-frontend/src/lib/modules/execution/components/schedule-overview/cross-execution-dashboard/cross-execution-dashboard.component.ts
```diff
@@ -18,12 +19,25 @@ import {
...
 
+  protected readonly nextExecutionDate$ = toObservable(this._state.task).pipe(
+    map((task) => ({ id: task?.id, isActive: task?.active !== false })),
+    distinctUntilChanged((previous, current) => previous.id === current.id && previous.isActive === current.isActive),
+    switchMap(({ id, isActive }) =>
+      id && isActive
+        ? this._schedulerService.getNextExecutionDate(id).pipe(catchError(() => of(undefined)))
+        : of(undefined),
+    ),
+  );
```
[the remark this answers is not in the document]
neogucky: Partially agree with this, but it doesn't solve all issues with this code, the next run is fetched only on initial task load and it becomes stale after executions. We should refresh it with the dashboard refresh flow.

projects/step-core/src/lib/client/generated/models/AsyncTaskStatus.ts
neogucky, line 1: Can you check if the new generated models are exported from generated/index.ts? I think something is missing, did you run the generation script for it?

neogucky (changes requested):

projects/step-core/src/lib/client/generated/models/AsyncTaskStatus.ts
lucianbaciu: @neogucky I had to manually remove something from index.ts - I think something is missconfigured. Having a master version, running the generate script should not throw errors normally.

neogucky: @lucian-baciu I do not understand this comment, did the script throw errors, and if so what where those?

lucianbaciu: @neogucky I had both FE and BE on most recent master. I ran node generate-openapi-client.js http://localhost:8080/rest/private-openapi.json, and the issues appeared. there were 2 double exports, but not in the same index.ts.
image
these 2 thigs export the same AsyncTaskStatus and ResourcePreview models, and the final export is ambiguous

lucianbaciu: if I removed them fron index.ts, it all worked

neogucky (dismissed):

neogucky (approved):

PR #1425, opened 2026-07-17
SED-4794 feature to export analytics table dashlet data as csv

lucianbaciu:


neogucky (changes requested):
I don't see that we are using the files from the generated client. I would prefer to exclude them if not needed as it reduces the changes in this PR.

neogucky (approved):

PR #1437, opened 2026-08-13
SED-3470 native federation

dvladir:


angular.json
neogucky: build:app:local seems to requests a non-existend local configuration, I guess we can remove it

projects/step-frontend/src/lib/plugins-initializer/plugins-initializer.ts
```diff
@@ -23,7 +23,11 @@ const fetchDefinitions = async (): Promise<MicrofrontendPluginDefinition[]> => {
       console.log('received plugins', plugin);
 
       if (plugin.entryPoint && plugin.entryPoint !== DEFAULT_ENTRY_POINT) {
-        plugin.entryPoint += '?v=${project.version}';
+        if (plugin.entryPoint.endsWith('.js')) {
+          plugin.entryPoint = plugin.entryPoint.replace('.js', '.json');
+        }
+        // TODO think about how to solve it properly. For now this string breaks the local run
+        // plugin.entryPoint += '?v=${project.version}';
```
neogucky: I had codex review the PR and it noted the following:

Plugin generation still uses the removed webpack federation package. The schematic invokes @angular-architects/module-federation, which is no longer installed after this migration, so generating a new plugin fails in a clean checkout.

package.json
neogucky: I guess we should also bump the required node version, what do you think?

neogucky (changes requested):

angular.json
dvladir: Missing configuration was added

projects/step-frontend/src/lib/plugins-initializer/plugins-initializer.ts
dvladir: Schematics has been removed

neogucky (dismissed):

